PR #1, opened 2017-12-13
Add singleton for virtualenv

nre-ableton:

### DIFF
--- a/vars/virtualenv.groovy
+++ b/vars/virtualenv.groovy
@@ -1,0 +1,14 @@
+def create(String python, String path = "venv", boolean installRequirements = true) {
+  sh("virtualenv -p ${python} ${path}")
+  if (installRequirements && fileExists("requirements.txt")) {
+    run(path:path, command:"pip install -r requirements.txt")
+  }
+}
+
+
+def run(String path = "venv", String command) {
+  sh("""
+    . ./${path}/bin/activate
+    ${command}
+  """)
+}


mst-ableton:
lgtm. do we have to worry about venv cleanup at all?

nre-ableton:
@mst-ableton I don't think it should be the responsibility of this library to do that, but I guess we could add a `cleanup` function.

mst-ableton:
what I was thinking was a context manager that wrapped `create` and deleted venv at the end of the context. but if it won't be used, then there's no point I guess?

nre-ableton:
@mst-ableton I see what you're getting at here, but I don't think it will work for a few reasons. Most notably, the `run` function takes a `String` and not a `body`. I would have preferred that approach for the same reasons as you so that we could write code like this:

```groovy
virtualenv.inside("python3.5") {
  sh "flake8 --whatever"
}
```

The problem is that this won't work. 😿 Since we need to `source` the `activate` script, we need to use a multi-line `sh` command, which would not work to execute `body.call()` inside of.

The other problem with this approach is that quite often want to use a venv in different stages of a build, and it is somewhat expensive to create. I wouldn't want the whole thing to be deleted and recreated every time the function was invoked.

mst-ableton:
What about something like this?

```
virtualenv.inside("python3.5") {  ctx ->
  ctx.run "flake8 --whatever"
  ctx.run "pylint --whatever"
}
```

where `ctx` has a `run()` method that prepends the activate command?

For the `venv` in disparate parts of the code issue, yeah the context manager wouldn't make sense. I think it completely makes sense to expose an unwrapped `create()` method in addition to the manager.

nre-ableton:
Yeah, I guess that approach could work, but semantically there is little difference between that and what's in the PR at the current state.

mst-ableton:
Ah you're right, the repeated `venv` creation would be small because it's already there. I guess you'd save on retyping python. In any case I'm not strongly in favor of it, and we can always add it later if we think we need it. lgtm as-is

vars/virtualenv.groovy
```diff
@@ -0,0 +1,14 @@
...
+  }
+}
+
+
+def run(String path = "venv", String command) {
+  sh("""
+    . ./${path}/bin/activate
+    ${command}
+  """)
+}
```
ala-ableton: @nre-ableton Please add an new line character at EOF.

nre-ableton: @ala-ableton will do!

vars/virtualenv.groovy
```diff
@@ -0,0 +1,14 @@
+def create(String python, String path = "venv", boolean installRequirements = true) {
```
ala-ableton: This function does too much to my taste. I think it should only do one thing: create a virtualenv. I am fine with adding a function to simplify calling `pip install -r requirements.txt`. but it doesn't belong in this one.

vars/virtualenv.groovy
```diff
@@ -0,0 +1,14 @@
...
+  sh("virtualenv -p ${python} ${path}")
```
ala-ableton: Let's use the long option `--python=${python}`.

ala-ableton: I don't like `path` being a default parameter, it's almost equivalent to hard-coding the value. Also I don't like `path` as a name, it's too generic. `virtualenv`'s help uses `dest_dir`.

vars/virtualenv.groovy
```diff
@@ -0,0 +1,14 @@
...
+  if (installRequirements && fileExists("requirements.txt")) {
+    run(path:path, command:"pip install -r requirements.txt")
+  }
+}
+
+
+def run(String path = "venv", String command) {
```
ala-ableton: I don't know how to interpret this... Default parameters should always be at the end, otherwise `run(foo)` and `run(foo, bar)` are not using `foo` the same way and :dizzy_face:

nre-ableton:
@ala-ableton Thanks for looking at this PR, I'll submit a new one to address your comments.

PR #3, opened 2017-12-13
Don't used named arguments when calling run

nre-ableton:
For some reason, Jenkins refuses to execute `run` when using named
arguments. There is no stacktrace and my groovy-fu is not good enough
to understand why it would fail. Regardless, the call also works fine
without named arguments.


ala-ableton:
LGTM :man_shrugging:

PR #5, opened 2017-12-14
Introduce VirtualEnv class

nre-ableton:
This commit introduces a class for VirtualEnv which keeps an internal
reference to the virtualenv destination dir. Also note that this
directory is no longer configurable, but instead is safely tucked away
in a temporary directory.

The virtualenv singleton has also been changed to remove all functions
other than create(), which now returns a new instance of the
VirtualEnv class.

### DIFF
--- a/src/com/ableton/VirtualEnv.groovy
+++ b/src/com/ableton/VirtualEnv.groovy
@@ -1,0 +1,26 @@
+package com.ableton
+
+
+class VirtualEnv implements Serializable {
+  def script
+  String destDir
+
+
+  VirtualEnv(script, String python) {
+    this.script = script
+    final PATH_SEP = script.isUnix() ? "/" : "\\"
+    this.destDir = script.pwd(tmp:true) +
+      PATH_SEP +
+      script.env.BUILD_NUMBER +
+      PATH_SEP +
+      python
+  }
+
+
+  def run(String command) {
+    script.sh("""
+      . ${destDir}/bin/activate
+      ${command}
+    """)
+  }
+}
--- a/vars/virtualenv.groovy
+++ b/vars/virtualenv.groovy
@@ -1,16 +1,8 @@
-def create(String python, String destDir) {
-  sh("virtualenv --python=${python} ${destDir}")
-}
+import com.ableton.VirtualEnv
 
 
-def installRequirements(String destDir) {
-  run(destDir, "pip install -r requirements.txt")
-}
-
-
-def run(String destDir, String command) {
-  sh("""
-    . ./${destDir}/bin/activate
-    ${command}
-  """)
+def create(script, String python) {
+  venv = new VirtualEnv(script, python)
+  venv.script.sh("virtualenv --python=${python} ${venv.destDir}")
+  return venv
 }


src/com/ableton/VirtualEnv.groovy
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+
+  def installRequirements() {
+    run("pip install -r requirements.txt")
+  }
+
+
+  def run(command) {
+    this.script.sh("""
```
ala-ableton: I am a bit confused about when we should prefix with `this.`. Always? Never? Depends? Or should we use a specific naming convention for member variables? Such as `mFoo` or `_foo`?

vars/virtualenv.groovy
```diff
@@ -1,16 +1,8 @@
...
 
 
-def run(String destDir, String command) {
-  sh("""
-    . ./${destDir}/bin/activate
-    ${command}
-  """)
+def create(script, String python) {
+  venv = new VirtualEnv(script)
+  venv.initialize(python)
```
ala-ableton: I think we should inline `initialize` here. Having it as a member function that could be called again feels dangerous.

src/com/ableton/VirtualEnv.groovy
```diff
@@ -0,0 +1,31 @@
...
+
+
+class VirtualEnv implements Serializable {
+  def script
+  def destDir
+
+  VirtualEnv(script) {
+    this.script = script
+    def pathSeparator = script.isUnix() ? "/" : "\\"
+    this.destDir = script.pwd(tmp:true) + pathSeparator + script.env.BUILD_NUMBER
```
ala-ableton: We should maybe add `python` to the `destDir`, so a given job can "compare" two Python versions. However, if we consider that a given job can create only one venv, then we should simply use a singleton.

vars/virtualenv.groovy
nre-ableton: I wish we could, but calling `shell.sh` throws a weird Jenkins exception. I can try again and give you a full stacktrace if you want more details.

src/com/ableton/VirtualEnv.groovy
nre-ableton: Yes, you're right, `this` isn't strictly necessary here. I'll remove it.

Regarding style... 😬 Yeah, at some point I would like to introduce a Groovy style guide, because the Java coder in me really hates the way our code looks. But if we are following standard Java conventions, there should be no prefix for fields or anything like that. It's only Google's wacky Java standards which recommend that.

src/com/ableton/VirtualEnv.groovy
nre-ableton: I really like this idea, if for no other reason than we'd see the python version in the logs for any `sh` calls.

vars/virtualenv.groovy
ala-ableton: What about `venv.script.sh(...)`?

nre-ableton: Ah, sorry, I misread your comment. The exception is thrown if we try calling `script.sh` from inside of the `VirtualEnv` ctor. Yeah, we could (and perhaps should) inline it here. Will fix.

src/com/ableton/VirtualEnv.groovy
```diff
@@ -5,10 +5,14 @@ class VirtualEnv implements Serializable {
   def script
   def destDir
 
-  VirtualEnv(script) {
+  VirtualEnv(script, python) {
     this.script = script
     def pathSeparator = script.isUnix() ? "/" : "\\"
-    this.destDir = script.pwd(tmp:true) + pathSeparator + script.env.BUILD_NUMBER
+    this.destDir = script.pwd(tmp:true) +
```
nre-ableton: FYI @ala-ableton, we should actually be putting `+` on the end of lines, not at the beginning of lines. Because Groovy doesn't have a line ending character (like `;`), having `+` at the beginning of the next line can cause some weird syntax errors.

src/com/ableton/VirtualEnv.groovy
```diff
@@ -0,0 +1,30 @@
...
+
+
+  def installRequirements() {
+    run("pip install -r requirements.txt")
+  }
+
+
+  def run(command) {
+    script.sh("""
+      . ${destDir}/bin/activate
```
ala-ableton: Can we use `source` instead? How does it work on Windows?

nre-ableton: Nope, `source` is a internal shell command but is technically an alias for `.` (which I learned when googling for how to source a script in Jenkins). So this is the correct way to do this.

It will work on Windows assuming that `sh` works at all on Windows. For the record, there is also a `bat` call for that platform.

src/com/ableton/VirtualEnv.groovy
```diff
@@ -0,0 +1,30 @@
+package com.ableton
+
+
+class VirtualEnv implements Serializable {
+  def script
+  def destDir
+
+  VirtualEnv(script, python) {
```
ala-ableton: `String python`?

src/com/ableton/VirtualEnv.groovy
```diff
@@ -0,0 +1,30 @@
...
+    this.script = script
+    def pathSeparator = script.isUnix() ? "/" : "\\"
```
ala-ableton: `final PATH_SEP`?

src/com/ableton/VirtualEnv.groovy
```diff
@@ -0,0 +1,30 @@
...
+    this.destDir = script.pwd(tmp:true) +
+      pathSeparator +
+      script.env.BUILD_NUMBER +
+      pathSeparator +
+      python
+  }
+
+
+  def installRequirements() {
```
ala-ableton: I think we don't need that function. 
```
venv.installRequirements()
```
vs
```
venv.run("pip install -r requirements.txt")
```
The second version is not that longer than the first one, but way more explicit.

src/com/ableton/VirtualEnv.groovy
ala-ableton: Sounds good.

src/com/ableton/VirtualEnv.groovy
```diff
@@ -0,0 +1,30 @@
...
+    run("pip install -r requirements.txt")
+  }
+
+
+  def run(command) {
```
ala-ableton: `String command`?

src/com/ableton/VirtualEnv.groovy
nre-ableton: Fixed. Also, I figured out the type of `script` by running `print this.getClass()` in the sandbox, and gave it a proper type as well.

src/com/ableton/VirtualEnv.groovy
nre-ableton: Done

src/com/ableton/VirtualEnv.groovy
nre-ableton: Ok, I removed this method.

src/com/ableton/VirtualEnv.groovy
nre-ableton: Done

src/com/ableton/VirtualEnv.groovy
ala-ableton: Nice!

vars/virtualenv.groovy
```diff
@@ -1,16 +1,8 @@
...
-}
+import com.ableton.VirtualEnv
 
 
-def run(String destDir, String command) {
-  sh("""
-    . ./${destDir}/bin/activate
-    ${command}
-  """)
+def create(script, String python) {
```
ala-ableton: `WorkflowScript script` then :wink:

nre-ableton: Ok sadly using a type for `WorkflowScript` doesn't work (see http://newci.office.ableton.com:8080/job/AbletonDevTools-runthebuilds/48/console). So we're back to just `def` for script. 😿

vars/virtualenv.groovy
```diff
@@ -1,16 +1,8 @@
...
+  venv = new VirtualEnv(script, python)
+  script.sh("virtualenv --python=${python} ${venv.destDir}")
```
ala-ableton: Last comment :smile:
`venv.script.sh()`?

nre-ableton: Technically it doesn't matter, but sure, I can do that.

ala-ableton:
LGTM :+1:

PR #6, opened 2018-01-03
Add unit tests, Gradle build

nre-ableton:
This PR adds a very minimal Gradle build script and some unit tests to this project. With the combination of these things, one can easily develop (and debug!) this project with an IDE such as IntelliJ and also run the tests from a local command line. Being able to run the unit tests outside of Jenkins itself has many advantages, but the JenkinsPipelineUnit library requires us to mock any Jenkins-specific functions that this library calls.

To that end, a `JenkinsMocks` class (and corresponding tests) is also contained in this branch. In the future, it might be nice to move this functionality to another library (a regular groovy library though, not a Jenkins pipeline library) which would be linked at compile time by projects such as this one. Alternatively, we might consider trying to merge them upstream to the JenkinsPipelineUnit project.

Also, it's worth mentioning the reason that there are no unit tests for the `run()` function of this library. The way that Jenkins executes commands via `sh` and the way they are executed by a local Java runtime (ie, the one invoked by Gradle) is a bit different. Specifically, Jenkins' runtime allows one to source bash scripts with the `.` command, which doesn't seem to work when run locally via `gradle test` (`java.lang.IOException` is thrown, complaining that it can't execute `.`).

This PR also contains one change to the API of the library, which is the introduction of the `create()` method in the `VirtualEnv` class. The reasoning behind this refactoring is explained in the message for this commit.


build.gradle
```diff
@@ -0,0 +1,30 @@
+apply plugin: 'groovy'
+
+repositories {
+  mavenCentral()
+}
+
+dependencies {
+  compile 'org.codehaus.groovy:groovy-all:2.4.10'
```
mst-ableton: does this automatically pull in groovy, or do we have to install it somewhere beforehand?

Jenkinsfile
```diff
@@ -13,6 +13,10 @@ def addStages() {
       codenarc: {
         codenarc.check('**/Jenkinsfile,**/*.gradle,**/*.groovy')
       },
+      junit: {
+        sh './gradlew test'
+        junit 'build/test-results/**/*.xml'
```
mst-ableton: where does `junit` come from?

build.gradle
nre-ableton: @mst-ableton Unless I'm mistaken, it automatically pulls it in. We don't have Groovy installed on the build nodes and the build worked there.

Jenkinsfile
nre-ableton: The JUnit Jenkins plugin, which is bundled with Jenkins... or maybe it's not bundled, but in that case it's a dependency of some other plugin we use. Either way, it's automatically installed whenever we deploy a new Jenkins.

test/com/ableton/JenkinsMocks.groovy
```diff
@@ -9,4 +9,12 @@ class JenkinsMocks {
   static Closure isUnix = {
     return !System.properties['os.name'].toLowerCase().contains('windows')
   }
+
+  static Closure pwd = { args ->
+    if (args?.tmp ?: false) {
```
mst-ableton: I'm not sure I know groovy enough to know what ?. and ?: mean. Does this mean if args.tmp is false? If so, why do we return 'java.io.tmpdir' here?

test/com/ableton/JenkinsMocks.groovy
```diff
@@ -17,4 +17,41 @@ class JenkinsMocks {
...
+      returnStatus = args['returnStatus'] ?: false
+      returnStdout = args['returnStdout'] ?: false
+      if (returnStatus && returnStdout) {
+        throw new IllegalArgumentException(
+          'returnStatus and returnStdout are mutually exclusive options')
+      }
+    }
+    assert script
+
+    Process process = script.execute()
```
mst-ableton: doesn't sound like a mock... sounds like it's actually running what we tell it to?

test/com/ableton/JenkinsMocks.groovy
ala-ableton: `a?.b` is the short version for `a == null ? null : a.b`  and `a ?: b` is the short version for `a ? a : b`.

test/com/ableton/JenkinsMocks.groovy
mst-ableton: unless you're saying the override in https://github.com/AbletonDevTools/python-pipeline-utils/pull/6/commits/9359777093ff166454fe8ac710983db7554d763f applies here?

test/com/ableton/VirtualEnvTest.groovy
```diff
@@ -0,0 +1,24 @@
...
+import static org.junit.Assert.assertNotNull
+
+
+/**
+ * Tests for the VirtualEnv class.
+ */
+class VirtualEnvTest extends BasePipelineTest {
+  def script
+
+
```
mst-ableton: not sure about groovy style, but should this only be one space?

ala-ableton: +1

src/com/ableton/VirtualEnv.groovy
```diff
@@ -12,7 +12,10 @@ class VirtualEnv implements Serializable {
 
   VirtualEnv(script, String python) {
     this.script = script
+    assert script
```
mst-ableton: is this here for debugging, or is it meant for production code? I know for python, raw `asserts` are seen as less-good in prod code and better error reporting is usually desired. is that the same for groovy?

src/com/ableton/VirtualEnv.groovy
```diff
@@ -7,6 +7,7 @@ package com.ableton
  */
 class VirtualEnv implements Serializable {
   def script
+  def python
```
mst-ableton: sort above `def script`?

src/com/ableton/VirtualEnv.groovy
mst-ableton: also, can we move both `assert script` and `assert python` to the top of the function? it looks like they're checking the results of a computation, when they're really checking the inputs

mst-ableton:
This generally seems fine to me, I mostly have style and readability concerns.

test/com/ableton/JenkinsMocks.groovy
nre-ableton: @mst-ableton In case you want to read more, `?.` is formally called the [safe navigation operator](http://groovy-lang.org/operators.html#_safe_navigation_operator) and `?:` is called the [elvis operator](http://groovy-lang.org/operators.html#_elvis_operator).

test/com/ableton/JenkinsMocks.groovy
nre-ableton: @mst-ableton No, that's correct... it's actually executing the script. It is possible to mock the output of `sh` by returning a string, but I found this to be rather clumsy especially when functions which contained multiple `sh` calls. I guess it's less of a "mock" and more of an "implementation" then. 😅 

I guess we could theoretically mock `sh` in a way that we pass it a map of key/value pairs corresponding to the arguments/mocked return value, but I don't like the direction that is heading in, since it would involve a lot of extra hacks in the `setUp` function.

test/com/ableton/VirtualEnvTest.groovy
nre-ableton: I'm fine with this being one space (actually I would prefer it to be so). We should probably make a section in our Groovy style guide then about vertical whitespacing in that case. In any event, I'll fix this.

src/com/ableton/VirtualEnv.groovy
nre-ableton: I'm not sure about Groovy code in general, but at least with respect to Jenkins pipelines I recall reading somewhere (though I can't find the link) that `assert` is a good practice since it's relatively cheap and will cause a build to fail for the right reasons.

In any case, for a shared library I think it's almost always appropriate to sanity check inputs since you have no idea how the code is being called.

nre-ableton: Also I agree that the `assert` statements should be together, I'll push a fixup for that.

src/com/ableton/VirtualEnv.groovy
nre-ableton: I would prefer that the order of declaration matches the order in which they are required in the constructor.

test/com/ableton/JenkinsMocks.groovy
mst-ableton: Thanks for the clarification. In that case I'm not sure why the `?:` is needed here. Let's say `q = args?.tmp`. Then we're doing `if (q ?: false) {`.  In that case, if `q` evaluates to `true` the elvis operator will return `q` (which, again, evaluates to `true`) and if `q` evaluates to false the elvis operator will return `false`. Since the results go directly into an if statement, I'm not sure what extra the elvis operator accomplishes.

test/com/ableton/JenkinsMocks.groovy
mst-ableton: The key/value pairs are what I typically do, it does mean extra work but it also means the tests can work on any OS.

I don't want to block this PR because of that, but we should at least make a comment here describing that the mock is only a semi-mock.

src/com/ableton/VirtualEnv.groovy
mst-ableton: 👍

src/com/ableton/VirtualEnv.groovy
```diff
@@ -25,6 +27,11 @@ class VirtualEnv implements Serializable {
   }
 
 
+  def create() {
+    script.sh("virtualenv --python=${python} ${destDir}")
+  }
+
+
```
mst-ableton: Unclear about the spacing here, it's 2-per-def here and 1-per-@test below. Assuming this is standard groovy, just wanted to check.

test/com/ableton/JenkinsMocks.groovy
nre-ableton: I guess you're right, I'll simplify this statement.

test/com/ableton/JenkinsMocks.groovy
nre-ableton: Nah, I'd rather get it right the first try rather than procrastinate with a `// TODO`. 😉 I think I have a good way to implement this, I'll push a fixup soon.

src/com/ableton/VirtualEnv.groovy
nre-ableton: Standard Groovy doesn't make any recommendations on vertical whitespacing. [Google's Java style guide](https://google.github.io/styleguide/javaguide.html#s4.6.1-vertical-whitespace) says:

> Multiple consecutive blank lines are permitted, but never required (or encouraged).

@ala-ableton encouraged the adoption of 2 newlines between methods, which I guess matches our C++ style guidelines. I would personally rather just go with 1 newline everywhere, but actually, this discussion should move to https://github.com/AbletonAppDev/guides/pull/55 instead. Let's sort out the style issues there, and then go through our Groovy code afterwards to make it compliant (which we'll pretty much have to do anyways).

test/com/ableton/JenkinsMocksTest.groovy
```diff
@@ -13,4 +14,12 @@ class JenkinsMocksTest extends BasePipelineTest {
   void setUp() {
     super.setUp()
   }
+
+
+  @Test
+  void isUnix() throws Exception {
```
ala-ableton: The function is called `isUnix()` but doesn't return a boolean...

ala-ableton: The function starts with `is` but doesn't return a boolean...

test/com/ableton/JenkinsMocksTest.groovy
```diff
@@ -13,4 +14,12 @@ class JenkinsMocksTest extends BasePipelineTest {
...
+    // It would be pretty redundant to basically re-implement this method in its own test
+    // case, so instead we just assert the call to see that it didn't go haywire.
+    assert JenkinsMocks.isUnix()
```
ala-ableton: What will happen if I run the tests on Windows?

test/com/ableton/JenkinsMocks.groovy
```diff
@@ -9,4 +9,12 @@ class JenkinsMocks {
   static Closure isUnix = {
     return !System.properties['os.name'].toLowerCase().contains('windows')
   }
+
+  static Closure pwd = { args ->
+    if (args?.tmp) {
+      return System.properties['java.io.tmpdir']
+    } else {
```
ala-ableton: I am not a fan of `return` after `else`.

test/com/ableton/JenkinsMocks.groovy
```diff
@@ -17,4 +17,74 @@ class JenkinsMocks {
...
+    int exitValue
+
+    MockScriptOutput(String stdout, int exitValue) {
+      this.stdout = stdout
+      this.exitValue = exitValue
+    }
+  }
+
+  /** Holds configured mock output values for the `sh` command. */
+  static Map<String, MockScriptOutput> mockScripts = [:]
```
ala-ableton: Rename to `mockScriptOutputs`?

test/com/ableton/JenkinsMocks.groovy
```diff
@@ -17,4 +17,74 @@ class JenkinsMocks {
...
+
+  /**
+   * Configure mock output for the `sh` command. This function should be called before
+   * attempting to call JenkinsMocks.sh.
```
ala-ableton: Please put backticks around `JenkinsMocks.sh` (and maybe also add `()` after `sh`).

test/com/ableton/JenkinsMocks.groovy
```diff
@@ -17,4 +17,74 @@ class JenkinsMocks {
...
+   * @see JenkinsMocks#sh
+   * @param script Script command to mock.
+   * @param stdout Standard output text to return for the given command.
+   * @param exitValue Exit value for the command (default: 0).
+   * @return
+   */
+  static addShMock(String script, String stdout, int exitValue = 0) {
```
ala-ableton: `static void`?

ala-ableton: Please remove the default value for `exitValue`.

test/com/ableton/JenkinsMocks.groovy
```diff
@@ -17,4 +17,74 @@ class JenkinsMocks {
...
+    mockScripts[script] = new MockScriptOutput(null, stdout, exitValue)
```
ala-ableton: Why are we passing `null` as the first argument here?

test/com/ableton/JenkinsMocks.groovy
```diff
@@ -17,4 +17,74 @@ class JenkinsMocks {
...
+  }
+
+  static Closure sh = { args ->
+    String script = null
+    def returnStdout = false
+    def returnStatus = false
+
+    // The `sh` function can be called with either a string, or a map of key/value pairs.
+    if (args instanceof String || args instanceof GString) {
```
ala-ableton: Out of curiosity, what the hell are `GString`s?

test/com/ableton/JenkinsMocks.groovy
```diff
@@ -17,4 +17,74 @@ class JenkinsMocks {
...
+    MockScriptOutput output = mockScripts[script]
+    if (!output) {
+      throw new IllegalArgumentException("No mock output configured for script call " +
+        "'${script}', did you forget to call JenkinsMocks.addShMock()?")
+    }
+    if (!returnStdout) {
+      println output.stdout
+    }
+    if (output.exitValue != 0) {
+      throw new Exception('Mock script failed with error code: ' + output.exitValue)
```
ala-ableton: If we throw when `output.exitValue != 0`, then we can always return 0 at line 85. So I think we shouldn't throw, right?

test/com/ableton/VirtualEnvTest.groovy
```diff
@@ -0,0 +1,23 @@
...
+ */
+class VirtualEnvTest extends BasePipelineTest {
+  def script
+
+  @Override
+  @Before
+  void setUp() throws Exception {
+    super.setUp()
+
+    this.script = loadScript('test/resources/EmptyPipeline.groovy')
```
ala-ableton: Doesn't this depend on the working directory when running the tests? Isn't there a more robust solution (based on file paths)?

test/com/ableton/VirtualEnvTest.groovy
```diff
@@ -19,5 +21,30 @@ class VirtualEnvTest extends BasePipelineTest {
...
+  }
+
+
+  @Test
+  void newObject() throws Exception {
+    def venv = new VirtualEnv(script, 'python2.7')
+    assertNotNull(venv)
+    assertNotNull(venv.script)
+    assertNotNull(venv.destDir)
+  }
```
ala-ableton: That test will only pass if I do have `python2.7` on my machine, right?

src/com/ableton/VirtualEnv.groovy
```diff
@@ -25,6 +27,11 @@ class VirtualEnv implements Serializable {
   }
 
 
+  def create() {
```
ala-ableton: I am against this change. However, I can propose the following alternative: make `create` a `static` method which does
```
  venv = new VirtualEnv(script, python)
  venv.script.sh("virtualenv --python=${python} ${venv.destDir}")
  return venv
```
and then call `VirtualEnv.create(script, python)` in the singleton `create`.

src/com/ableton/VirtualEnv.groovy
ala-ableton: The rule I wanted to adopt is the Python rule. 2 newlines between top-level things (classes, functions, ...) and 1 newline between nested things (methods, ...). So there should be only 1 empty line in this case.

nre-ableton:
@ala-ableton I know it seems weird to have the Gradle wrapper + jars in the repo, but it is a best practice when using Gradle. This makes it possible for developers (or build systems) to be able to build the software using a guaranteed-compatible version of Gradle without having to have Gradle itself installed on their system. See:

https://stackoverflow.com/questions/20348451/why-should-the-gradle-wrapper-be-committed-to-vcs
https://stackoverflow.com/questions/21762180/should-you-include-or-ignore-gradle-wrapper-properties

src/com/ableton/VirtualEnv.groovy
nre-ableton: Ok, I'll fix this (and the other parts of the code) accordingly, and also update the style guide with that guideline.

src/com/ableton/VirtualEnv.groovy
nre-ableton: Ok, I could live with that.

test/com/ableton/JenkinsMocks.groovy
nre-ableton: Actually we can refactor to this one-liner:

```groovy
    return args?.tmp ? System.properties['java.io.tmpdir'] : System.properties['user.dir']
```

test/com/ableton/JenkinsMocks.groovy
nre-ableton: Sure, sounds good.

test/com/ableton/JenkinsMocks.groovy
nre-ableton: Done.

test/com/ableton/JenkinsMocks.groovy
ala-ableton: Of course :sweat_smile:

ala-ableton: Actually: `return System.properties[args.?tmp ? 'java.io.tmpdir' : 'user.dir']

test/com/ableton/JenkinsMocks.groovy
nre-ableton: Done.

test/com/ableton/JenkinsMocks.groovy
nre-ableton: When you create an inner class in Groovy, the ctor gets implicit first argument for the outer class type. Unless we pass `null` here, Groovy will complain that no method matches the signature.

test/com/ableton/JenkinsMocks.groovy
nre-ableton: When you define a string with `'hello'`, it's a `java.lang.String`, but if you want to use interpolation (ie, `"hello ${thing}"`), then it's a Groovy GString. See:

http://mrhaki.blogspot.se/2009/08/groovy-goodness-string-strings-strings.html

test/com/ableton/JenkinsMocks.groovy
nre-ableton: This is meant to emulate Jenkins' behavior for `sh`. In a Jenkins pipeline, `sh` will throw if the exit code is anything other than 0.

test/com/ableton/VirtualEnvTest.groovy
nre-ableton: Gradle will always run the tests using the root directory of the project as the working directory.

test/com/ableton/JenkinsMocks.groovy
ala-ableton: Thanks for the clarification!

test/com/ableton/VirtualEnvTest.groovy
ala-ableton: Good to know!

test/com/ableton/JenkinsMocksTest.groovy
nre-ableton: This conforms to JUnit naming guidelines. We are testing the `isUnix` method, and the `@Test` annotation tells us that. Previous to the introduction of that annotation (which was added in JUnit 4.0, IIRC), this would be declared as `void testIsUnix() throws Exception`, which ended up producing very verbose and redundant test code. 😉 Since we are using `@Test` it is not necessary to name this method `@Test void testIsUnix()`.

Also, JUnit test methods are required to return `void`.

test/com/ableton/JenkinsMocksTest.groovy
nre-ableton: Ahh, good catch. I'll refactor.

test/com/ableton/JenkinsMocksTest.groovy
ala-ableton: All good then!

test/com/ableton/VirtualEnvTest.groovy
nre-ableton: Actually the tests will fail everywhere because I didn't update `VirtualEnvTest` yet. 😅 I wanted to push my new implementation of `sh` with the mocked output the script return values, but I'll fix this now.

nre-ableton:
@ala-ableton I think I've addressed all of your comments, mind taking another look?

ala-ableton:
:warning: 621b89ca6c8b9a2a37ca5e5e81267a4f93fe8925 won't apply cleanly

test/com/ableton/JenkinsMocksTest.groovy
```diff
@@ -29,7 +29,7 @@ class JenkinsMocksTest extends BasePipelineTest {
   void isUnix() throws Exception {
     // It would be pretty redundant to basically re-implement this method in its own test
     // case, so instead we just assert the call to see that it didn't go haywire.
-    assert JenkinsMocks.isUnix()
+    JenkinsMocks.isUnix()
```
ala-ableton: @nre-ableton The comment should also change

test/com/ableton/JenkinsMocks.groovy
ala-ableton: Then I don't understand how `returnStatus` works...

test/com/ableton/JenkinsMocks.groovy
ala-ableton: Thanks!

src/com/ableton/VirtualEnv.groovy
ala-ableton: Do you intend to address my comment?

ala-ableton:
@nre-ableton It seems that one of my comments didn't get addressed (making `VirtualEnv.create` static), and I made another comment about a comment :smile:. Please rebase and apply the fixup commits before I do another review round.

src/com/ableton/VirtualEnv.groovy
nre-ableton: I thought I did? I made the `create` method in `VirtualEnv` static, and fixed the tests accordingly. Is there some miscommunication going on here?

ala-ableton: Sorry. I didn't realize this was the only commit that was re-written.

nre-ableton:
@ala-ableton Ok, rebased and squashed!

test/com/ableton/JenkinsMocksTest.groovy
```diff
@@ -13,4 +14,11 @@ class JenkinsMocksTest extends BasePipelineTest {
   void setUp() {
     super.setUp()
   }
+
+  @Test
+  void isUnix() throws Exception {
+    // It would be pretty redundant to basically re-implement this method in its own test
+    // case, so instead we just assert the call to see that it didn't go haywire.
```
ala-ableton: @nre-ableton we don't assert anymore.

test/com/ableton/VirtualEnvTest.groovy
ala-ableton: Is this something you still want to do in this PR?

src/com/ableton/VirtualEnv.groovy
ala-ableton: That member is not needed anymore.

ala-ableton:
@nre-ableton I am done with my new review round. I still don't understand the point of `returnStatus` if `sh` throws when the exit code is non-zero...

test/com/ableton/JenkinsMocksTest.groovy
nre-ableton: Sorry, I forgot to fix this in my last round.

nre-ableton:
@ala-ableton I was wrong about the behavior of `sh`, I just ran some tests in the sandbox. If a script returns a non-zero code, it should return that value when `returnStatus` is true, otherwise `sh` returns `false`. I will fix accordingly.

src/com/ableton/VirtualEnv.groovy
nre-ableton: True, fixed.

nre-ableton:
@ala-ableton Thanks! 🎉 

Hmm, it looks like CI wasn't configured to automatically build PRs/merges for this repo, I would like to force-push again just to have the assurance that the tests pass (even though I have also run them locally).

nre-ableton:
Hmm, it looks like codenarc 0.2.0 has some issues 😅 I will fix that separately.

PR #8, opened 2018-01-04
Use version 0.2.0 of codenarc library

nre-ableton:
ptal @AbletonDevTools/gotham-city , thanks!


nre-ableton:
@rco-ableton The reason the cleanups are included in this commit is that otherwise we'd end up with at least one commit that would break the build (if built atomically, which we don't do, but whatever). That is, if this branch had the cleanup commit first, it would fail with the codenarc 0.1.0 configuration, and if we updated to codenarc 0.2.0 first, that commit would be broken without these cleanups.

rco-ableton:
@nre-ableton okay, cool, thanks for the explanation!

PR #11, opened 2018-01-11
Update to codenarc 0.5.0

nre-ableton:
Related to https://github.com/AbletonAppDev/devtools/issues/856, ptal @AbletonDevTools/gotham-city, thanks!

### DIFF
--- a/test/com/ableton/VirtualEnvTest.groovy
+++ b/test/com/ableton/VirtualEnvTest.groovy
@@ -1,12 +1,12 @@
 package com.ableton
 
-import com.lesfurets.jenkins.unit.BasePipelineTest
-import org.junit.Before
-import org.junit.Test
-
 import static org.junit.Assert.assertEquals
 import static org.junit.Assert.assertNotNull
 import static org.junit.Assert.fail
+
+import com.lesfurets.jenkins.unit.BasePipelineTest
+import org.junit.Before
+import org.junit.Test
 
 
 /**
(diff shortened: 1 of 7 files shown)


test/com/ableton/VirtualEnvTest.groovy
```diff
@@ -40,7 +40,8 @@ class VirtualEnvTest extends BasePipelineTest {
   @Test
   void newObjectWithNullScript() throws Exception {
     try {
-      def venv = new VirtualEnv(null, 'python2.7')
+      @SuppressWarnings('UnusedVariable')
+      def unused = new VirtualEnv(null, 'python2.7')
```
ala-ableton: Why not `new VirtualEnv(null, 'python2.7')` only?

test/com/ableton/VirtualEnvTest.groovy
```diff
@@ -50,13 +51,15 @@ class VirtualEnvTest extends BasePipelineTest {
   @Test
   void newObjectWithNullPython() throws Exception {
     try {
-      def venv = new VirtualEnv(script, null)
+      @SuppressWarnings('UnusedVariable')
+      def unused = new VirtualEnv(script, null)
```
ala-ableton: Why not `new VirtualEnv(script, null)` only?

test/com/ableton/VirtualEnvTest.groovy
```diff
@@ -50,13 +51,15 @@ class VirtualEnvTest extends BasePipelineTest {
...
       fail('Expected exception, but none was thrown')
     } catch (AssertionError error) {
       assertNotNull(error)
     }
   }
 
+  @SuppressWarnings('BuilderMethodWithSideEffects')
```
ala-ableton: Could this be moved closer to the line that triggers the warning?

test/com/ableton/VirtualEnvTest.groovy
nre-ableton: Hmm, good point. I had avoided doing that because my IDE would throw a warning with `Result of VirtualEnv() is ignored`, but it seems that codenarc doesn't care about this so I guess it is fine. 😅 

(Same goes for the same comment in other lines)

test/com/ableton/VirtualEnvTest.groovy
nre-ableton: Yes, it can be moved 1 line down, I guess.

ala-ableton:
LGTM :+1:

PR #14, opened 2018-01-18
Test cases should not call fail in a try block

nre-ableton:
This exception will be caught, which can lead to false positives. ping @AbletonDevTools/gotham-city


ala-ableton:
LGTM 👍

PR #17, opened 2018-01-29
Introduce new singleton/class to install Python

nre-ableton:
The new PythonBuilder class will download a Python tarball from the
internet, extract it to the job's temporary directory, compile it, and
then return a string to the installation directory (also located in
the job's temporary directory). This takes about two minutes on one of
our single-core Linux nodes, but can be sped up with a number of
obvious optimizations.

The idea is to be able to test against new releases of python for
which packages are not yet available.

### DIFF
--- a/src/com/ableton/PythonBuilder.groovy
+++ b/src/com/ableton/PythonBuilder.groovy
@@ -1,0 +1,101 @@
+package com.ableton
+
+
+class PythonBuilder implements Serializable {
+  @SuppressWarnings('FieldTypeRequired')
+  def script = null
+  /**
+   * Python major version to download (required value). Example: '3.7.0'
+   */
+  String version = null
+
+  /**
+   * Mirror to use when downloading Python.
+   */
+  String downloadMirror = 'https://www.python.org/ftp/python'
+  /**
+   * Method to use to download Python. Currently supported methods include:
+   * - 'curl': Requires curl to be installed on the build node.
+   * - 'httpRequest': Requires the Jenkins HTTP Request Plugin to be installed (default).
+   */
+  String downloadWith = 'httpRequest'
+  /**
+   * Number of parallel jobs to use when building Python (ie, `make -jN`). Default: '1'
+   */
+  String makeJobs = '1'
+  /**
+   * Python version revision, optional. This value is appended to the major version, but
+   * because of the way that Python's downloads are organized, this value must be separate
+   * from the major version. So for Python 3.7.0a4, use '3.7.0' for the version and 'a4'
+   * for the revision.
+   */
+  String revision = ''
+
+  String install() {
+    // Sanity check required values
+    assert script
+    assert version
+
+    if (!script.isUnix()) {
+      throw new UnsupportedOperationException('This function is only supported on Unix')
+    }
+
+    String sourcesTarball = download()
+    String sourcesPath = extract(sourcesTarball)
+    return build(sourcesPath)
+  }
+
+  protected String pythonBaseName() {
+    return 'Python-' + version + revision
+  }
+
+  protected String tempDir() {
+    return script.pwd(tmp: true)
+  }
+
+  protected String download() {
+    String downloadUrl = downloadMirror + '/' + version +
+      '/' + pythonBaseName() + '.tgz'
+    String outputFile = script.pwd(tmp: true) + '/' + pythonBaseName() + '.tgz'
+    script.echo 'Downloading python from: ' + downloadUrl
+
+    switch (downloadWith) {
+      case 'curl':
+        script.sh "curl -s -o ${outputFile} ${downloadUrl}"
+        break
+      case 'httpRequest':
+        script.httpRequest(outputFile: outputFile, url: downloadUrl)
+        break
+      default:
+        script.error 'Unknown downloadWith type: ' + downloadWith
+    }
+
+    return outputFile
+  }
+
+  protected String extract(String sourcesTarball) {
+    script.echo 'Extracting python sources'
+    script.dir(tempDir()) {
+      script.sh 'tar xfz ' + sourcesTarball
+    }
+    return tempDir() + '/' + pythonBaseName()
+  }
+
+  protected String build(String sourcesPath) {
+    // Ensure that makeJobs is not an empty string, since make is a bit dumb and running
+    // `make -j` will spawn one job for every file, potentially forkbombing the node.
+    if (!makeJobs || !makeJobs.toInteger()) {
+      throw new IllegalArgumentException('Invalid argument for makeJobs: ' + makeJobs)
+    }
+
+    script.echo 'Building python from: ' + sourcesPath
+    String installPath = script.pwd(tmp: true) + '/' + pythonBaseName()
+    script.dir(sourcesPath) {
+      script.sh './configure --prefix=' + installPath
+      script.sh 'make -j' + makeJobs
+      script.sh 'make install'
+      script.deleteDir()
+    }
+    return installPath
+  }
+}
--- a/test/com/ableton/PythonBuilderTest.groovy
+++ b/test/com/ableton/PythonBuilderTest.groovy
@@ -1,0 +1,102 @@
+package com.ableton
+
+import static org.junit.Assert.assertEquals
+import static org.junit.Assert.assertNotNull
+
+import com.lesfurets.jenkins.unit.BasePipelineTest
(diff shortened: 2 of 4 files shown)


nre-ableton:
Well, I didn't get any feedback on my initial API design, but I polished it up anyways and added test cases. ptal @AbletonDevTools/gotham-city, thanks!

ala-ableton:
I am not fond of the naming. `installPython.run()` doesn't match `PythonBuilder().installI()`.

nre-ableton:
@ala-ableton Should we go with `pythonBuilder.install()`?

ala-ableton:
If we can, sure!

ala-ableton:
The commit message mentions `PythonBuilderTest` instead of `PythonBuilder`.

src/com/ableton/PythonBuilder.groovy
```diff
@@ -0,0 +1,101 @@
...
+
+
+class PythonBuilder implements Serializable {
+  @SuppressWarnings('FieldTypeRequired')
+  def script = null
+  /**
+   * Python major version to download (required value). Example: '3.7.0'
+   */
+  String version = null
+
```
ala-ableton: Style nitpicking, why is there an empty here?

nre-ableton:
@ala-ableton Fixed, also I totally forgot to add the newly-created test cases to that commit (also fixed).

src/com/ableton/PythonBuilder.groovy
```diff
@@ -0,0 +1,101 @@
...
+  String revision = ''
+
+  String install() {
+    // Sanity check required values
+    assert script
+    assert version
+
+    if (!script.isUnix()) {
+      throw new UnsupportedOperationException('This function is only supported on Unix')
+    }
```
ala-ableton: What is only supported on Unix?
- Calling `download()`? Then `download()` itself should complain.
- Calling `extract()`? Then `extract()` itself should complain.
- Calling `build()`? Then `build()` itself should complain.

src/com/ableton/PythonBuilder.groovy
nre-ableton: I wanted to differentiate between the required and optional parameters.

ala-ableton: I see. Then let's leave it here.

src/com/ableton/PythonBuilder.groovy
nre-ableton: All of those functions will only work on Unix. 😄 This is mostly due to assumptions about path separators, but there are other things which could easily go wrong on Windows which I have not tested. For example, we call `sh` instead of `bat` which won't work on Windows hosts which lack a Unix toolchain. Our Windows build hosts have that, but we can't make any such assumptions for others (and this is a public library).

I would prefer to exclude that platform until we have a need to support it, rather than invest the time in making this code cross-platform without any real need to do so.

Am I being too lazy here? 😅

ala-ableton: You are not. I am fine with excluding Windows. But let's exclude it properly.

nre-ableton: How should we do that, then?

ala-ableton: Can we add a ctor and throw in the ctor?

nre-ableton: No, I would prefer not to do that. I have really come around to the idea of not defining ctors and letting Groovy autogenerate them, ever since we decided to go that route with `DevToolsProject`.

Let me explain the way the language works, because it might not be obvious here (it wasn't to me until I read up on it). By not defining a ctor but instead setting default values for fields, you can use "named arguments" when creating an object like so:

```groovy
class Foo {
  String bar = 'bar'
  String baz = 'baz'
}

Foo foo = new Foo(bar: 'not-bar', baz: 'not-baz')
```

I think that we can both agree that we like this syntax. 😁 So the way that works behind the scenes is that the Groovy generates a bunch of default ctors, including one which takes a `Map` and (presumably) uses all sorts of reflection to assign the values to their respective fields. The call to `new Foo` converts the args to a map behind the scenes (note that we *don't* need to use `[:]` when calling `new Foo([bar: 'not-bar' ...])`). So the point is, once you have a custom ctor, you can't create objects using the named argument syntax. In other words, this will break:

```groovy
class Foo {
  String bar
  String baz

  Foo(String bar, String baz) {
    this.bar = bar
    this.baz = baz
  }
}

// Error: Will throw groovy.lang.GroovyRuntimeException
// With a message like "Could not find matching constructor for: Foo(java.util.LinkedHashMap)"
Foo foo = new Foo(bar: 'not-bar', baz: 'not-baz')
```

So basically, named parameters aren't really named parameters. As such, I'd prefer to use the calling syntax where we can pass named parameters rather than bloat the ctor and `install` with a bunch of arguments with default values.

nre-ableton: Hmm, now that I think about it, I guess we *could* also provide a custom ctor which also takes a `Map`, but I find this code to be rather ugly:

```groovy
class Foo {
  String bar
  String baz

  Foo(Map args) {
    bar = args?.bar ?: 'bar'
    baz = args?.baz ?: 'baz'
  }
}
```

Regardless, not generating the ctor and letting Groovy do its thing is definitely the "Groovy way".

ala-ableton: Then let's not add a ctor :+1: 
I've just realized that `install` is the only public method :sweat_smile: Forget about what I said, sorry for the noise :retreat:

nre-ableton: @ala-ableton Sounds good, what changes do I need to make to this PR then? I think I've addressed your other concerns, no?

test/com/ableton/PythonBuilderTest.groovy
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @Test(expected = Exception)
+  void installWithBuildError() throws Exception {
+    PythonBuilder pb = new PythonBuilder(script: script, version: '1.0.0')
+    String pythonPath = pb.tempDir() + '/' + pb.pythonBaseName()
+    JenkinsMocks.addShMock('tar xfz ' + pythonPath + '.tgz', '', 0)
+    JenkinsMocks.addShMock('./configure --prefix=' + pythonPath, '', 0)
+    JenkinsMocks.addShMock('make -j1', '', 1)
+    JenkinsMocks.addShMock('make install', '', 0)
```
ala-ableton: Is it really necessary to add that sh mock for `make install`?

test/com/ableton/PythonBuilderTest.groovy
```diff
@@ -0,0 +1,103 @@
...
+
+  @Test(expected = IllegalArgumentException)
+  void installWithEmptyMakeJobs() throws Exception {
+    PythonBuilder pb = new PythonBuilder(script: script, version: '1', makeJobs: '')
+    String pythonPath = pb.tempDir() + '/' + pb.pythonBaseName()
+    JenkinsMocks.addShMock('tar xfz ' + pythonPath + '.tgz', '', 0)
+    pb.install()
+  }
+
+  @Test(expected = NumberFormatException)
```
ala-ableton: What is throwing this exception?

test/com/ableton/PythonBuilderTest.groovy
nre-ableton: No, it's not, so I will remove this. FYI this was necessary with jenkins-pipeline-mocks 0.2.0 (which I started out with when I wrote these tests), but in 0.4.1 we fixed `sh`'s exception throwing behavior so it is no longer necessary.

test/com/ableton/PythonBuilderTest.groovy
nre-ableton: The `toInteger()` call in `build()`.

nre-ableton:
FYI @ala-ableton, @rco-ableton, here's a preview of what mergeomatic's Jenkinsfile will look like with this: https://github.com/AbletonDevTools/mergeomatic/pull/74

It fails on Jenkins because we will need some extra software installed on our Linux nodes to build Python 3.7, namely `zlib1-dev` and a few other packages which currently escape me. 😅

test/com/ableton/PythonBuilderTest.groovy
ala-ableton: I see.

PR #18, opened 2018-01-31
Use separate build directory when building Python

nre-ableton:
Previous to this commit, we were using the same path for building the
sources as was passed to `./configure --prefix`. Not only is this
generally a bad idea, but we also call `deleteDir()` after building,
which would remove the newly built Python installation.


nre-ableton:
re-ping @AbletonDevTools/gotham-city

ala-ableton:
LGTM :+1:

PR #21, opened 2018-02-02
Use singleton for script context

nre-ableton:
We don't need to pass a script variable to singletons, because the
singleton has access to execute script commands. Therefore the
singleton's own `this` can be passed to the VirtualEnv ctor here.


ala-ableton:
Nice! LGTM :+1:

PR #23, opened 2018-02-12
Add script argument in pythonBuilder singleton

nre-ableton:


ala-ableton:
As usual, why?

nre-ableton:
@ala-ableton So that callers don't need to add `script: this` when calling this singleton. I also added this information to the commit message body.

ala-ableton:
LGTM :+1:

PR #29, opened 2018-03-20
Use system temporary directory

nre-ableton:
Some Jenkins job types, namely multi-branch pipelines created by the
GitHub Organization jobs, have very long workspace directory names.
This can cause problems when using virtualenv since the script shebang
is limited to 128 characters.

---


ala-ableton:
These changes look good, but I would like to see them in action first.

nre-ableton:
Related to https://github.com/AbletonAppDev/devtools/issues/1025

nre-ableton:
@ala-ableton So the approach that you reviewed doesn't work, because `dir` is CPS-transformed and hence can't be called from a constructor (see https://issues.jenkins-ci.org/browse/JENKINS-26313, https://github.com/jenkinsci/workflow-cps-plugin/blob/master/README.md#technical-design, for more explanation). So we're back to using `/tmp`. 😿

ala-ableton:
a05c5e5c03a13aea3bbac2e8890c58432b14eef0 should be introduced before 70253d26707a54d43bf7b76365d2edbfb9a21fd2, otherwise there is no way to clean the /tmp folder. We could even add that function on a separate PR (which then could bump to `0.8.2`).

70253d26707a54d43bf7b76365d2edbfb9a21fd2's commit message contains `WIP: tmp dir`
11e70ba0fda27b785dcb4cfa0c799adae91ae7a7 should be squashed into 70253d26707a54d43bf7b76365d2edbfb9a21fd2.

nre-ableton:
@ala-ableton fixed!

ala-ableton:
LGTM :+1:

PR #33, opened 2018-08-16
Check only *.gradle files in top-level directory

nre-ableton:


mst-ableton:
@nre-ableton is there a bug describing why we're doing this?

PR #38, opened 2018-10-10
Introduce new class: Pipenv

nre-ableton:
This class is designed to facilitate running a Pipenv command against
multiple versions of Python, which is something that Pipenv cannot do
by itself very gracefully. We do this by using the `--python` argument
to Pipenv, which assumes that the `Pipfile` does not contain a
`requirements` section which may cause Pipenv to throw an error.

---

This is the first step in solving https://github.com/AbletonAppDev/devtools/issues/1340, ping @AbletonDevTools/gotham-city


nre-ableton:
re-ping @AbletonDevTools/gotham-city

src/com/ableton/Pipenv.groovy
```diff
@@ -0,0 +1,114 @@
...
+    assert script
+    assert args
+    assert args.command
+    assert args.pythonVersions
+
+    Map result = [:]
+
+    try {
+      args.pythonVersions.each { python ->
+        script.sh "pipenv install --dev --python ${python}"
```
ala-ableton: It should be documented somewhere that we are passing `--dev` to `pipenv install`.

src/com/ableton/Pipenv.groovy
```diff
@@ -0,0 +1,114 @@
...
+   *                 <strong>Note:</strong> this option and the {@code returnStatus}
+   *                 option are mutually exclusive.
+   *               </li>
+   *             </ul>
+   * @return Map of return values. The keys in the map correspond to the Python versions
+   *         given in {@code args.pythonVersions}, and the values are either integers,
+   *         strings, or null, depending on whether {@code returnStatus},
+   *         {@code returnStdout}, or neither option was specified (respectively).
+   */
+  Map runCommand(Map args) {
```
ala-ableton: Why do we need both `runCommand` and `runWith`? `runCommand` seems to be a subset of `runWith`.

src/com/ableton/Pipenv.groovy
```diff
@@ -0,0 +1,114 @@
...
+  Map runWith(Map args, Closure body) {
+    assert script
+    assert args
+    assert args.pythonVersions
+
+    Map result = [:]
+
+    try {
+      args.pythonVersions.each { python ->
+        script.sh "pipenv install --dev --python ${python}"
```
ala-ableton: It should be documented somewhere that we are passing `--dev` to `pipenv install`.

src/com/ableton/Pipenv.groovy
nre-ableton: It's a convenience, but I'm open to removing `runCommand` and only having `runWith`.

src/com/ableton/Pipenv.groovy
nre-ableton: Fixed

src/com/ableton/Pipenv.groovy
nre-ableton: I've removed this method (see below)

vars/pipenv.groovy
```diff
@@ -0,0 +1,24 @@
...
+
+
+/**
+ * Run a command with Pipenv using multiple versions of Python.
+ * @param args Map of arguments.
+ * @return Map with return output or values for each Python version.
+ * @see com.ableton.Pipenv#runCommand(Map)
+ */
+Map runCommand(Map args) {
+  return new Pipenv(script: this).runCommand(args)
```
ala-ableton: @nre-ableton this should be removed.

nre-ableton: @ala-ableton Good catch! Removed.

src/com/ableton/Pipenv.groovy
```diff
@@ -0,0 +1,52 @@
...
+   * parallelized and therefore the commands are run serially for each Python version.
+   *
+   * This function also installs the development packages for each Python version (in
+   * other words, it runs {@code pipenv install --dev}. Also, it removes the virtualenv
+   * after the last Python version has been run.
+   *
+   * @param pythonVersions List of Python versions to run the command with. This argument
+   *                       is passed to Pipenv via {@code pipenv --python}. See
+   *                       {@code pipenv --help} for supported syntax.
+   * @param body Closure body to execute. The closure body is passed the python version
```
ala-ableton: Nitpick: `python version` -> `Python version`

src/com/ableton/Pipenv.groovy
```diff
@@ -0,0 +1,52 @@
...
+   *             as an argument.
+   * @return Map of return values. The keys in the map correspond to the Python versions
+   *         given in {@code args.pythonVersions}, and the values are the result of
```
ala-ableton: `are the results`?

src/com/ableton/Pipenv.groovy
```diff
@@ -0,0 +1,52 @@
...
+   *         executing the closure body.
+   */
+  Map runWith(List pythonVersions, Closure body) {
+    assert script
+    assert pythonVersions
+
+    Map result = [:]
```
ala-ableton: `result` -> `results`?

src/com/ableton/Pipenv.groovy
nre-ableton: Fixed, and also improved the surrounding grammar.

nre-ableton:
@ala-ableton Regarding the green PR on another repo: https://github.com/AbletonDevTools/ablbot/pull/186

I'm going to go ahead and merge this change, assuming there are no objections.

PR #39, opened 2018-11-08
Add PythonPackageVersionReader class

mst-ableton:
This moves https://github.com/AbletonAppDev/devtools/issues/1433 forward.

@ala-ableton @nre-ableton @rco-ableton @eta-ableton ptal!


src/com/ableton/PythonPackageVersionReader.groovy
```diff
@@ -0,0 +1,56 @@
+package com.ableton
+
+
+/**
+ * Class to get the version of a Python package by parsing the file that sets it.
+ */
+class PythonPackageVersionReader implements Serializable {
```
nre-ableton: Given that the library is called `python-utils`, I think the class could simply be called `PackageVersionReader`. What do you think about this?

src/com/ableton/PythonPackageVersionReader.groovy
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * End of the line setting the Python package version.
+   */
+  String versionLineEnding = "'"
+
+  /**
+   * Read a Python file setting the version of a package (usually {@code __init__.py})
+   * and return its version as a {@code String}.
+   * @param filename The file to read and parse the version out of.
+                     <strong>Required value, may not be null!</strong>
```
nre-ableton: I don't think this is necessary to document for parameters. Unless a default value is provided, it's usually assumed that they are required. We only assert parameters as a sanity check.

src/com/ableton/PythonPackageVersionReader.groovy
```diff
@@ -0,0 +1,56 @@
...
+    assert versionLineEnding
+    assert filename
+
+    String fileContents = script.readFile(filename).trim()
+
+    String version
+    fileContents.split('\n').each { line ->
+      if (line.startsWith(versionLineBeginning)) {
+        // Extract the version string from the quotes
+        version = line.replace(
```
nre-ableton: Does this not fit on 90 chars?

test/com/ableton/PythonPackageVersionReaderTest.groovy
```diff
@@ -0,0 +1,85 @@
...
+  @SuppressWarnings('FieldTypeRequired')
+  def script
+
+  @Override
+  @Before
+  void setUp() throws Exception {
+    super.setUp()
+
+    this.script = loadScript('test/resources/EmptyPipeline.groovy')
+    assertNotNull(script)
```
nre-ableton: I don't thing this check is really necessary (though it isn't hurting anything either).

test/com/ableton/PythonPackageVersionReaderTest.groovy
```diff
@@ -0,0 +1,85 @@
...
+
+    helper.registerAllowedMethod('readFile', [String], JenkinsMocks.readFile)
+
+    JenkinsMocks.addReadFileMock('__init__.py', '''
+# Copyright (c) 2018 Ableton AG, Berlin. All rights reserved.\n
+\n
+"""gruberflask is designed to help you obtain $640 million in bearer bonds."""\n
```
nre-ableton: Cute 😸

test/com/ableton/PythonPackageVersionReaderTest.groovy
```diff
@@ -0,0 +1,85 @@
...
+\n
+__version__ = '1.2.3'\n
+''',
+    )
+  }
+
+  @Test
+  void getPythonPackageVersion() throws Exception {
+    String version = new PythonPackageVersionReader(
```
nre-ableton: Please break on the `.` operator, ie:

```groovy
new Foo(script: script)
  .bar('blah')
```

src/com/ableton/PythonPackageVersionReader.groovy
```diff
@@ -0,0 +1,56 @@
...
+   * Script context.
+   * <strong>Required value, may not be null!</strong>
+   */
+  @SuppressWarnings('FieldTypeRequired')
+  def script = null
+
+  /**
+   * Start of the line setting the Python package version.
+   */
+  String versionLineBeginning = "__version__ = '"
```
nre-ableton: I'm not sure I like `'` being included in this string, and for that matter, the presence of the `versionLineEnding` operator below. Might it make more sense to just define this as `versionLineBeginning = '__version__'` and then have `boolean stringsAreSingleQuoted = true`? Then `=` is also somewhat assumed.

src/com/ableton/PythonPackageVersionReader.groovy
mst-ableton: I was concerned with the resulting singleton -- suddenly you have `packageVersionReader` which could be any kind of language package...

nre-ableton: We can always name the singleton something else.

ala-ableton: But don't we take the risk of having some naming clash? For instance, what if I were to create a `PackageVersionReader` in `rust-utils` and then import both `python-utils` and `rust-utils`?

src/com/ableton/PythonPackageVersionReader.groovy
mst-ableton: Fixed as of 2437eb2.

test/com/ableton/PythonPackageVersionReaderTest.groovy
mst-ableton: Fixed as of 2437eb2.

test/com/ableton/PythonPackageVersionReaderTest.groovy
mst-ableton: Fixed as of 2437eb2.

src/com/ableton/PythonPackageVersionReader.groovy
mst-ableton: @nre-ableton how about renaming it to `PythonPackage`, with a method `readVersion()`? That way it's nice and short, and we avoid collisions. In the future, we might be able to use `PythonPackage` to manipulate or get other kinds of information from the package. What do you think?

mst-ableton: This has been addressed in bdba347.

src/com/ableton/PythonPackage.groovy
```diff
@@ -0,0 +1,67 @@
...
+    String versionLineBeginning = args.versionLineBeginning ?: defaultVersionLineBeginning
+    String versionLineEnding = args.versionLineEnding ?: defaultVersionLineEnding
+
+    String fileContents = script.readFile(args.filename).trim()
+
+    String version
+    fileContents.split('\n').each { line ->
+      if (line.startsWith(versionLineBeginning)) {
+        // Extract the version string from the quotes
+        version = line.replace(versionLineBeginning, '').replace(versionLineEnding, '')
```
ala-ableton: Shouldn't we `break` once we have what we want?

src/com/ableton/PythonPackage.groovy
```diff
@@ -0,0 +1,67 @@
...
+    assert script
+    assert args.filename
+    String versionLineBeginning = args.versionLineBeginning ?: defaultVersionLineBeginning
+    String versionLineEnding = args.versionLineEnding ?: defaultVersionLineEnding
+
+    String fileContents = script.readFile(args.filename).trim()
+
+    String version
+    fileContents.split('\n').each { line ->
+      if (line.startsWith(versionLineBeginning)) {
```
ala-ableton: Why are we checking only for `startsWith(versionLineBeginning)`? If the line is `__version__ = '0.1.0'  # lol`, then the version will be `0.1.0  # lol`.

mst-ableton: Fair enough!

src/com/ableton/PythonPackage.groovy
mst-ableton: https://stackoverflow.com/questions/3049790/can-you-break-from-a-groovy-each-closure

ala-ableton: Let's use `find` then. We are actually looking for a particular line.

src/com/ableton/PythonPackage.groovy
```diff
@@ -0,0 +1,67 @@
...
+    String versionLineBeginning = args.versionLineBeginning ?: defaultVersionLineBeginning
+    String versionLineEnding = args.versionLineEnding ?: defaultVersionLineEnding
+
+    String fileContents = script.readFile(args.filename).trim()
+
+    String version
+    fileContents.split('\n').each { line ->
+      if (line.startsWith(versionLineBeginning) && line.endsWith(versionLineEnding)) {
+        // Extract the version string from the quotes
+        version = line.replace(versionLineBeginning, '').replace(versionLineEnding, '')
```
ala-ableton: Using `replace` is potentially dangerous if `versionLineBeginning` and/or `versionLineEnding` also appear at another location of the line. I think the following would be safer:
```groovy
        version = line.substring(
            versionLineBeginning.length(),
            line.length() - versionLineEnding.length(),
        )
```

mst-ableton: Sounds good. please reformat at a GitHub suggestion or a fixup commit

ala-ableton:
```suggestion
        version = line.substring(
            versionLineBeginning.length(),
            line.length() - versionLineEnding.length(),
        )
```

mst-ableton: CodeNarc disallows `substring()`. I've replaced this with string slicing in 530b7ae. I'm not sure how we handle Groovy string slicing formatting, if this formatting doesn't fit our style please submit a GitHub suggestion to correct it.

src/com/ableton/PythonPackage.groovy
mst-ableton: Addressed in 530b7ae.

mst-ableton:
@ala-ableton how does 530b7ae look to you?

mst-ableton:
@nre-ableton what say ye?

PR #48, opened 2019-08-14
Make VirtualEnv class, tests work on Windows

nre-ableton:


src/com/ableton/VirtualEnv.groovy
```diff
@@ -50,15 +52,14 @@ class VirtualEnv implements Serializable {
...
-      pathSep +
-      script.env.JOB_BASE_NAME +
-      pathSep +
-      script.env.BUILD_NUMBER +
-      pathSep +
-      python.split(pathSep).last()
+    String tempDir = '/tmp'
+    if (!script.isUnix()) {
+      List tempDirParts = script.env.TEMP.split(':')
+      tempDir = "/${tempDirParts[0]}${tempDirParts[1].replace('\\', '/')}"
```
mst-ableton: Don't we have to prepend like `/cygdrive/` or something?

nre-ableton: No, what I meant is that the paths need to look like the ones used by cygwin when you use the bash shell. So `C:\foo` becomes `/c/foo`, and so on.

PR #54, opened 2020-01-30
Remove PythonBuilder class

nre-ableton:
This is untested and unused. We have no way of knowing whether it still
works, so best not to support it further.


nre-ableton:
@ablbot rebase

PR #55, opened 2020-01-30
Use Object type instead of def + suppression

nre-ableton:


nre-ableton:
@dgl-ableton We want to use `Object` instead of `def`, not the other way around. 😄 The reason for this is that CodeNarc (aka groovylint) warns about `def`, so it requires a suppression. `Object` is safe for non-primitive types and requires no such suppression, so that has become a recommended practice for our Groovy code.

PR #56, opened 2020-02-10
Allow Python versions to be loaded via pyenv

rco-ableton:
Overloads the create() method of VirtualEnv to allow passing a new
parameter pointing to the root of the pyenv installation. If present,
a virtualenv will be created using the Python version specified by the
"python" parameter. This string should match a valid Python version as
it would be passed to "pyenv install".


ala-ableton:
@rco-ableton I'm interested in reviewing this PR, thanks!

dgl-ableton:
@ablbot autosquash

vars/virtualenv.groovy
```diff
@@ -1,11 +1,21 @@
...
 
 
+/**
+ * Create a new {@link com.ableton.VirtualEnv} instance.
+ * @param python Python version or absolute path to Python executable.
+ * @param pyenvRoot Path of the pyenv installation.
+ * @return New {@link com.ableton.VirtualEnv} instance.
+ */
+VirtualEnv create(String python, String pyenvRoot) {
+  return VirtualEnv.create(script, python, pyenvRoot)
```
dgl-ableton: Where is `script` coming from?

vars/virtualenv.groovy
```diff
@@ -1,11 +1,21 @@
...
+}
+
 /**
  * Create a new {@link com.ableton.VirtualEnv} instance.
  * @param python Python version or absolute path to Python executable.
  * @return New {@link com.ableton.VirtualEnv} instance.
  */
 VirtualEnv create(String python) {
-  return VirtualEnv.create(this, python)
+  return VirtualEnv.create(python)
```
dgl-ableton: Is this calling this https://github.com/AbletonDevTools/python-pipeline-utils/pull/56/commits/b0f1c3a783c3f162c1198b46d0c83a326f80a86f#diff-2424724bbebf6d79ee5c8cdb6edc3e0dR38?

src/com/ableton/VirtualEnv.groovy
```diff
@@ -15,12 +15,59 @@ class VirtualEnv implements Serializable {
...
+   * variables (e.g. PYENV_ROOT and PATH) will be done inside this function.
+   *
+   * @param script Script context.
+   *               <strong>Required value, may not be null!</strong>
+   * @param python Python version, as given by pyenv versions --list.
+   * @param pyenvRoot Path to the installation of pyenv.
+   * @return New instance of VirtualEnv object.
+   */
+  static VirtualEnv create(Object script, String python, String pyenvRoot) {
+    if (!script.isUnix()) {
```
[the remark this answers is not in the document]
nre-ableton: It's part of the Jenkins pipeline API, and thus part of the script context. `script` itself is not an `Object`, but it's technically a `WorkflowScript` object (which of course ultimately derives from `Object`). However, we can't call it a `WorkflowScript` because we don't have the library for that symbol in our project. Hence, we just type it as a generic `Object`.

src/com/ableton/VirtualEnv.groovy
```diff
@@ -15,12 +15,59 @@ class VirtualEnv implements Serializable {
...
+      script.error 'This method is not supported on Windows'
+    }
 
-  protected String activateSubDir
+    VirtualEnv venv = new VirtualEnv(script, python)
+    assert pyenvRoot
+
+    if (!script.fileExists(pyenvRoot)) {
```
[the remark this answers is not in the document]
nre-ableton: I agree that this is nice, however @rco-ableton should be warned that this will break the tests and will require updating the `@Test(expected = ...)` parts, since we'll be raising only `Exception` now and not `AssertionError`.

vars/virtualenv.groovy
nre-ableton: Good catch @dgl-ableton! It should be:

```suggestion
  return VirtualEnv.create(this, python, pyenvRoot)
```

vars/virtualenv.groovy
nre-ableton: No, it's calling https://github.com/AbletonDevTools/python-pipeline-utils/commit/b0f1c3a783c3f162c1198b46d0c83a326f80a86f#diff-2424724bbebf6d79ee5c8cdb6edc3e0dR43. However this line is broken and needs to be reverted.

nre-ableton:
Sorry for jumping in this PR @dgl-ableton, but since I co-authored this with @rco-ableton I thought I could answer some of your questions directly. 😄

src/com/ableton/VirtualEnv.groovy
```diff
@@ -15,12 +15,59 @@ class VirtualEnv implements Serializable {
...
+   * should be installed on the node, but the actual setup of any required environment
+   * variables (e.g. PYENV_ROOT and PATH) will be done inside this function.
+   *
+   * @param script Script context.
+   *               <strong>Required value, may not be null!</strong>
+   * @param python Python version, as given by pyenv versions --list.
+   * @param pyenvRoot Path to the installation of pyenv.
+   * @return New instance of VirtualEnv object.
+   */
+  static VirtualEnv create(Object script, String python, String pyenvRoot) {
```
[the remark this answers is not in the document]
rco-ableton: Where are you seeing that it's defined as a `String`?

src/com/ableton/VirtualEnv.groovy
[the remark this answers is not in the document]
rco-ableton: It would be nice but I think I'll leave this alone for now. Feel free to open a pull request updating it though!

src/com/ableton/VirtualEnv.groovy
```diff
@@ -15,12 +15,59 @@ class VirtualEnv implements Serializable {
...
+
+    venv.script.sh("""
+      ${venv.envSetup}
+      pyenv install -s ${python}
+      pyenv shell ${python}
+      pip install virtualenv
+      virtualenv ${venv.destDir}
+    """)
+
+    venv.envSetup += ". ${venv.destDir}/${venv.activateSubDir}/activate"
```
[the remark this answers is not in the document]
rco-ableton: This is a good point—I'll have to let @nre-ableton comment here because he wrote this part 🙂

src/com/ableton/VirtualEnv.groovy
```diff
@@ -15,12 +15,59 @@ class VirtualEnv implements Serializable {
    * Destination directory for the virtualenv. This value is set during construction of
    * the object, and is under the system temporary directory.
    */
-  String destDir
+  String destDir = null
+  /**
+   * Series of commands needed to install and active a version of Python inside the
```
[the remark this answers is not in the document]
rco-ableton: Addressed in https://github.com/AbletonDevTools/python-pipeline-utils/pull/56/commits/5dde137bc6fa4b88dc716ebd3015019290443784.

vars/virtualenv.groovy
rco-ableton: Addressed in https://github.com/AbletonDevTools/python-pipeline-utils/pull/56/commits/5dde137bc6fa4b88dc716ebd3015019290443784.

vars/virtualenv.groovy
rco-ableton: Addressed in https://github.com/AbletonDevTools/python-pipeline-utils/pull/56/commits/5dde137bc6fa4b88dc716ebd3015019290443784.

src/com/ableton/VirtualEnv.groovy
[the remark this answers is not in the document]
nre-ableton: @dgl-ableton I believe you are mistaken, but I like your paranoia. 😄  On Line 50 above (https://github.com/AbletonDevTools/python-pipeline-utils/pull/56/files#diff-2424724bbebf6d79ee5c8cdb6edc3e0dR50) we overwrite `venv.envSetup` by assignment. So this append operator should be safe.

src/com/ableton/VirtualEnv.groovy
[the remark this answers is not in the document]
dgl-ableton: Ah sorry I must have misread `Script` as `String` 😅

src/com/ableton/VirtualEnv.groovy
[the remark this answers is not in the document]
dgl-ableton: I see :D Maybe this is a silly question, but then why are we appending `venv.envSetup`?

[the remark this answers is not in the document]
nre-ableton: Because we must run the above `sh` call _without_ activating the venv. So what that script is doing is effectively:

1. Add environment variables needed for pyenv first
2. Add a call to `pyenv init -`
3. Use these values to ensure that the desired python version is installed
4. Install virtualenv itself
5. Create a virtualenv for the project

We can't add a call to `. ${venv.destDir}/${venv.activateSubDir}/activate` to `venv.envSetup` right away because the venv only gets created in steps 4-5. To avoid this chicken-and-egg problem, we just append the `activate` bit to `venv.envSetup` after we have created the venv itself.

nre-ableton:
@rco-ableton I'd like to also review this PR before it gets merged, if that's alright.

rco-ableton:
@ala-ableton all yours when you're ready 🙂

src/com/ableton/VirtualEnv.groovy
```diff
@@ -15,12 +15,59 @@ class VirtualEnv implements Serializable {
    * Destination directory for the virtualenv. This value is set during construction of
    * the object, and is under the system temporary directory.
    */
-  String destDir
+  String destDir = null
+  /**
+   * Series of commands needed to install and activate a version of Python inside the
+   * current shell.
+   */
+  String envSetup = null
```
ala-ableton:
```suggestion
  String activateCommands = null
```

src/com/ableton/VirtualEnv.groovy
```diff
@@ -15,12 +15,59 @@ class VirtualEnv implements Serializable {
...
+    """)
+
+    venv.envSetup += ". ${venv.destDir}/${venv.activateSubDir}/activate"
+
+    return venv
+  }
 
   /**
-   * Factory method to create new class instance and properly initialize it.
+   * Create a virtualenv using a specific locally installed version of Python.
```
ala-ableton: Nitpicking, this could be done in a separate commit.

test/com/ableton/VirtualEnvTest.groovy
```diff
@@ -125,6 +126,54 @@ class VirtualEnvTest extends BasePipelineTest {
...
   }
 
+  @SuppressWarnings('TrailingWhitespace')
+  @Test
+  void createPyenv() throws Exception {
+    String pythonVersion = '1.2.3'
+    String pyenvRoot = '/mock/pyenv/root'
+    helper.registerAllowedMethod('fileExists', [String]) { return true }
+    helper.registerAllowedMethod('isUnix', []) { return true }
+    // Note: The extra whitespace is needed here to match the sh mock
```
ala-ableton:
```suggestion
    // Note: The extra empty line is needed here to match the sh mock
```

src/com/ableton/VirtualEnv.groovy
```diff
@@ -15,12 +15,59 @@ class VirtualEnv implements Serializable {
...
+    VirtualEnv venv = new VirtualEnv(script, python)
+    assert pyenvRoot
+
+    if (!script.fileExists(pyenvRoot)) {
+      script.error "pyenv root path '${pyenvRoot}' does not exist"
+    }
+
+    venv.envSetup = """
+      export PYENV_ROOT=${pyenvRoot}
+      export PATH=\$PYENV_ROOT/bin:\$PATH
```
ala-ableton: Why isn't `PYENV_ROOT` set on the node when installing `pyenv`? And why doesn't installing `pyenv` add it to the `PATH`?

vars/virtualenv.groovy
```diff
@@ -1,6 +1,16 @@
 import com.ableton.VirtualEnv
 
 
+/**
+ * Create a new {@link com.ableton.VirtualEnv} instance.
+ * @param python Python version or absolute path to Python executable.
+ * @param pyenvRoot Path of the pyenv installation.
+ * @return New {@link com.ableton.VirtualEnv} instance.
+ */
+VirtualEnv create(String python, String pyenvRoot) {
```
ala-ableton: I think it would be nicer to have a `createWithPyenv` function that would only take the `python` version.

vars/virtualenv.groovy
```diff
@@ -1,6 +1,16 @@
 import com.ableton.VirtualEnv
 
 
+/**
+ * Create a new {@link com.ableton.VirtualEnv} instance.
+ * @param python Python version or absolute path to Python executable.
```
ala-ableton:
```suggestion
 * @param python Python version, as given by pyenv versions --list.
```
at least based on the documentation of `VirtualEnv.create`.

vars/virtualenv.groovy
ala-ableton: I guess we should bump the patch version to acknowledge the addition of this function and the other one in `VirtualEnv`.

src/com/ableton/VirtualEnv.groovy
nre-ableton: I like this suggestion!

src/com/ableton/VirtualEnv.groovy
nre-ableton: I'll jump in here for @rco-ableton -- it's because there isn't really a concept of a global environment on Unix, as there is on Windows.

ala-ableton: I didn't realize that each time we were doing PATH manipulation in cidre (for Python, rust, and CMake), it was always on Windows. 
However, we know and control the user that is running the jobs (`jenkins`), so we could set `PYENV_ROOT` and `PATH` for this user, right?

rco-ableton: We tried this, but it doesn't work—from memory I think it was something like Jenkins was erasing the user's environment and not picking up the `PYENV_ROOT` etc.

ala-ableton: Then the current solution is fine I guess.

vars/virtualenv.groovy
nre-ableton: Fixed in e29dd17f52ce9b759b342e37fcdab1aa79e9e62a (though as we discussed on the VC, not as you suggested above).

nre-ableton: Whoops, I nearly forgot to do that myself. Fixed in dd109cb.

nre-ableton:
FWIW @ala-ableton, I tested the singleton changes with this sandbox script:

```groovy
library 'python-utils@pyenv_support'

node('master') {
    try {
        // fails, because we don't configure pyenv on the master
        def v = virtualenv.createWithPyenv('3.8.0')
    } catch (ignored) {}
}

node('generic-linux') {
    // works!
    def v = virtualenv.createWithPyenv('3.8.0')
    // fails as expected ("ERROR: pyenv root path '/tmp/foo' does not exist")
    def v2 = virtualenv.createWithPyenv('3.8.0', '/tmp/foo')
}
```

ala-ableton:
@ablbot autosquash

test/com/ableton/VirtualEnvTest.groovy
```diff
@@ -125,6 +126,54 @@ class VirtualEnvTest extends BasePipelineTest {
...
+  @SuppressWarnings('TrailingWhitespace')
+  @Test
+  void createPyenv() throws Exception {
+    String pythonVersion = '1.2.3'
+    String pyenvRoot = '/mock/pyenv/root'
+    helper.registerAllowedMethod('fileExists', [String]) { return true }
+    helper.registerAllowedMethod('isUnix', []) { return true }
+    // Note: The extra empty line is needed here to match the sh mock
+    JenkinsMocks.addShMock("""
+
```
ala-ableton: I don't know whether it would work, but something like
```suggestion
      ${String()}
```
would prevent me from having to change the settings of my text editor wrt. trailing spaces when I save this file.
Also it would allow removing the warning suppression.

test/com/ableton/VirtualEnvTest.groovy
```diff
@@ -125,6 +126,54 @@ class VirtualEnvTest extends BasePipelineTest {
...
+    String pyenvRoot = '/mock/pyenv/root'
+    helper.registerAllowedMethod('fileExists', [String]) { return true }
+    helper.registerAllowedMethod('isUnix', []) { return true }
+    // Note: The extra empty line is needed here to match the sh mock
+    JenkinsMocks.addShMock("""
+      
+      export PYENV_ROOT=${pyenvRoot}
+      export PATH=\$PYENV_ROOT/bin:\$PATH
+      eval "\$(pyenv init -)"
+
```
ala-ableton:
```suggestion
    ${String()}
```
similarly to the previous comment.

test/com/ableton/VirtualEnvTest.groovy
nre-ableton: I'll give it a try!

nre-ableton: So `${String()}` doesn't work (same with a few variations thereof which I tried), but I created an empty string in a variable and expanded the comment, see faf5a41.

test/com/ableton/VirtualEnvTest.groovy
nre-ableton: See faf5a41.

test/com/ableton/VirtualEnvTest.groovy
```diff
@@ -125,6 +126,56 @@ class VirtualEnvTest extends BasePipelineTest {
     assertEquals(venv.destDir, createdVenv.destDir)
   }
 
+  @SuppressWarnings('TrailingWhitespace')
```
ala-ableton:
```suggestion
```

nre-ableton: D'oh!

nre-ableton:
@ala-ableton Thanks for the fixup!

nre-ableton:
@ablbot autosquash

PR #58, opened 2020-03-10
Create virtualenv in build temporary folder

ala-ableton:


nre-ableton:
I would like to review this PR after @rco-ableton, thanks!

nre-ableton:
So this LGTM, but my only concern is _why_ did we not use `pwd(tmp: true)` in the first place, and decide to use a hardcoded `/tmp` folder? I looked in the git history and couldn't figure this out. Was the idea to actually preserve the virtualenvs between builds, so we could improve caching?

Also, I am not 100% certain how Jenkins cleans up temporary directories. It would be useful to do some experiments on that. Is there some way that we can verify this by testing it against one of our projects?

ala-ableton:
@nre-ableton thanks a lot for sharing your concerns. I did a few experiments using a `sandbox` job and indeed Jenkins doesn't clean up any directory (temporary or not) between builds. I guess I need to go back to the drawing board 😅

nre-ableton:
@ala-ableton Sounds good. You could make another PR with your cleanup commits, though.

Also, I guess the question is whether the workspace cleanup option of `checkout` is sufficient. All things being equal, Jenkins doesn't clean up `/tmp` either, so maybe there's really no major difference here. Just something to think about...

PR #59, opened 2020-04-30
Update JenkinsPipelineUnit, jenkins-pipeline-mocks

nre-ableton:


build.gradle
```diff
@@ -2,6 +2,9 @@ apply plugin: 'groovy'
 
 repositories {
   mavenCentral()
+  maven {
```
rco-ableton, line 5: I don't understand why this is required for the version bump?

build.gradle
```diff
@@ -11,12 +14,13 @@ dependencies {
   compile 'com.github.AbletonDevTools:jenkins-pipeline-mocks:0.7+'
   compile 'org.codehaus.groovy:groovy-all:2.4.12'
 
+  testCompile 'org.assertj:assertj-core:3.15+'
```
rco-ableton, line 17: Same here, I don't get how this is related to updating the version number.

test/com/ableton/VirtualEnvTest.groovy
```diff
@@ -25,8 +25,6 @@ class VirtualEnvTest extends BasePipelineTest {
     assertNotNull(script)
     script.env = ['BUILD_NUMBER': 1, 'JOB_BASE_NAME': 'mock']
 
-    helper.registerAllowedMethod('deleteDir', [], JenkinsMocks.deleteDir)
```
rco-ableton, line 28: Are these lines no longer required in v0.8?

build.gradle
nre-ableton: JenkinsPipelineUnit is not being published to Maven Central anymore, just to the Jenkins repository. See https://github.com/jenkinsci/JenkinsPipelineUnit/issues/173 if you are really curious.

test/com/ableton/VirtualEnvTest.groovy
nre-ableton: Correct, they were removed in https://github.com/AbletonDevTools/jenkins-pipeline-mocks/pull/38 (which was actually reviewed by you 😉).

build.gradle
nre-ableton: TBH I also don't completely understand why this is necessary, but it seems to be related to how dependencies are managed by JenkinsPipelineUnit. I've filed an issue about that here: https://github.com/jenkinsci/JenkinsPipelineUnit/issues/209.

Regardless, this issue would be fixed in 1.5, so to bump to 1.4 we'd still need the `assertj-core` dependency.

nre-ableton:
@rco-ableton I added a bit more information in the commit messages.

build.gradle
rco-ableton: Understood—it would be worth adding a comment explaining and linking to the issue you filed, I think.

PR #62, opened 2020-08-03
Add override for run method for map argument

nre-ableton:
This PR improves the unit tests for the `VirtualEnv` class and adds support for
calling `sh` with the `Map` override, which will allow one to get the output or
status code from commands executed in the virtualenv.


ala-ableton:
Very nice!

PR #79, opened 2021-04-13
Store virtualenvs in workspace, not temp folder

nre-ableton:
Note that this may cause problems for users who are using this library
with the GitHub Organization plugin. Hence, it is a breaking API change.


src/com/ableton/VirtualEnv.groovy
```diff
@@ -104,18 +104,13 @@ class VirtualEnv implements Serializable {
...
     } else {
       activateSubDir = 'Scripts'
-      List tempDirParts = script.env.TEMP.split(':')
-      tempDir = "/${tempDirParts[0]}${tempDirParts[1].replace('\\', '/')}"
     }
 
-    this.destDir = "${tempDir}/${script.env.JOB_BASE_NAME}/${script.env.BUILD_NUMBER}/" +
-      python.split('/').last()
-
+    this.destDir = "${script.env.WORKSPACE}/.venv/${python.split('/').last()}"
```
ala-ableton, line 108: This is problematic. If a build on a given node needs 2 different venvs for the same Python version, in the best case there will be an error, in the worst case the existing venv will be reused and some bugs might sneak in.

The problem existed before, but it was countered by the `.cleanup()` method.

src/com/ableton/VirtualEnv.groovy
```diff
@@ -114,16 +114,6 @@ class VirtualEnv implements Serializable {
     this.activateCommands = ". ${destDir}/${activateSubDir}/activate"
   }
 
-  /**
-   * Removes the Virtualenv from disk. You should call this method in the cleanup stage
-   * of the pipeline to avoid cluttering the build node with temporary files.
-   */
-  void cleanup() {
```
ala-ableton, line 126: I think we should keep this method. It gives more flexibility wrt. how the VirtualEnv objects can be used. Also, it reduces the "breakingness" of this PR.

ala-ableton (commented):
I like where this is going, but I think we need to make more precise changes.

src/com/ableton/VirtualEnv.groovy
nre-ableton: Let's instead use a random number, similar to what we do in `postgres-utils` for the port. See 5089c97.

src/com/ableton/VirtualEnv.groovy
nre-ableton: Sure, I can see that. See 8c3d29d.

src/com/ableton/VirtualEnv.groovy
```diff
@@ -24,6 +26,13 @@ class VirtualEnv implements Serializable {
 
   protected String activateSubDir = null
 
+  /**
+   * Seed for random number generator. Dy default, a unique seed will be used based on the
```
ala-ableton, line 30:
```suggestion
   * Seed for random number generator. By default, a unique seed will be used based on the
```

src/com/ableton/VirtualEnv.groovy
```diff
@@ -74,11 +83,14 @@ class VirtualEnv implements Serializable {
    *
    * @param script Script context.
    *               <strong>Required value, may not be null!</strong>
-   * @param python Python version or absolute path to Python executable.
+   * @param python Python version or absolute path to Python executable. On Windows, this
+   *               should be a Cygwin-style path, but <strong>without the {@code .exe}
+   *               extension</strong>, for example: {@code /c/Python27/python}
```
ala-ableton, line 88: Why is the docstring for this parameter being extended in 5089c97e90f3b39e5622a88893b72a165b019fd3? Also, the documentation is not accurate, `create('python2.7')` and `create('python3.7')` work fine on Windows.

ala-ableton (commented):
Please rebase/squash and I'll have another look, thanks!

src/com/ableton/VirtualEnv.groovy
nre-ableton: I'll make another cleanup commit for the documentation.

nre-ableton: I had meant to put that in another cleanup commit, but forgot. I guess it's not adding much, though, so I'll remove it.

test/com/ableton/VirtualEnvTest.groovy
```diff
@@ -33,16 +33,13 @@ class VirtualEnvTest extends BasePipelineTest {
 
   @Test
   void cleanup() {
-    script.env['TEMP'] = 'C:\\Users\\whatever\\AppData\\Temp'
-
-    VirtualEnv venv = new VirtualEnv(script, 'python3.6')
+    VirtualEnv venv = new VirtualEnv(script)
```
ala-ableton, line 36: Removing the `python` argument belongs to e5190442246eea104e15ffa4a9f5d7648f203cc1, not  945f26ff9b108991a3c67cc1332143ab35997711.

src/com/ableton/VirtualEnv.groovy
```diff
@@ -96,11 +106,11 @@ class VirtualEnv implements Serializable {
    * @param python Python version or absolute path to Python executable. On Windows, this
    *               should be a Cygwin-style path, but <strong>without the {@code .exe}
    *               extension</strong>, for example: {@code /c/Python27/python}
+   * @param randomSeed If non-zero, use this seed for the random number generator.
    * @see #create(Object, String)
    */
-  VirtualEnv(Object script, String python) {
```
ala-ableton, line 101: In e5190442246eea104e15ffa4a9f5d7648f203cc1, the docstring for the `python` argument should be removed.

test/com/ableton/VirtualEnvTest.groovy
nre-ableton: This was a rebase error, I think. See 11d3416.

src/com/ableton/VirtualEnv.groovy
nre-ableton: Fixed in 47c5159.

test/com/ableton/VirtualEnvTest.groovy
```diff
@@ -177,7 +168,9 @@ class VirtualEnvTest extends BasePipelineTest {
     helper.addShMock(mockScriptCall, 'mock output', 0)
     helper.registerAllowedMethod('isUnix', []) { return true }
 
-    new VirtualEnv(script, 'python').run('mock-script')
+    VirtualEnv venv = new VirtualEnv(script)
+    venv.randomSeed = 1
+    venv.run('mock-script')
```
ala-ableton, line 173: Why is this split over 3 lines?

ala-ableton (commented):
Please rebase/squash before requesting another review, thanks!

test/com/ableton/VirtualEnvTest.groovy
nre-ableton: I think this was a remnant of a previous implementation of the random seed. Same goes for the field. I've fixed both things now.

ala-ableton (approved):

PR #88, opened 2021-10-11
Use a regexp to parse the version

nre-ableton:
As this is an API-breaking change, we will need to bump the minor version
number.

FYI @rco-ableton.


rco-ableton (approved):
LGTM 👍

PR #90, opened 2021-11-25
Init pyenv with --path

nre-ableton:
This fixes an issue where the system pip would be used instead of
pyenv's pip, which may or may not exist and also may not be the correct
version.


nre-ableton:
FYI @ala-ableton

ala-ableton:
@nre-ableton the README of https://github.com/pyenv/pyenv does say:

> If you are installing Pyenv yourself as part of the batch job, after installing the files, run the following in the job's shell to be able to use it.
> ```
> export PYENV_ROOT="$HOME/.pyenv"
> export PATH="$PYENV_ROOT/bin:$PATH"    # if `pyenv` is not already on PATH
> eval "$(pyenv init --path)"
> eval "$(pyenv init -)"
> ```

nre-ableton:
@ala-ableton Yep, see e495240. I've tested this internally and it works.

rco-ableton (approved):
LGTM 👍

ala-ableton:
@nre-ableton no the hard question. Does this still work with pyenv 1? If not, do we drop support for pyenv 1 or do we bump the version of `python-utils` to mark a breaking change?

nre-ableton:
@ala-ableton It will not work with pyenv v1.x. I verified this by cloning the pyenv git repo and running:

```
97f244fb!pyenv$ ./bin/pyenv init --path
# Load pyenv automatically by appending
# the following to your profile:

eval "$(pyenv init -)"

97f244fb!pyenv$ echo $?
1
```

I'd rather bump the version than make another `sh` call, to be honest.

nre-ableton:
Actually, `eval` will eat the return code, so it won't fail on pyenv 1.x. We don't need to bump the minor version.

nre-ableton:
@ablbot autosquash

PR #102, opened 2022-09-13
Refactor `VirtualEnv` class to only set `PATH`, remove old/untested classes

nre-ableton:
This PR contains two major API changes, and although normally I'd submit two PRs in this case, I felt it made more sense to submit a single PR since we'll need to make a minor version bump.

The changes are as follows:

1. Removal of the `pipenv` and `virtualenvs` singletons and associated classes.
2. Refactoring of the `VirtualEnv` class to no longer source the `activate` script.

---

### Removal of old classes

We don't use them anymore and aren't testing them in the "real world". In the former case (`pipenv`), we have no idea if this code still works with modern `pipenv` versions (have there been any?). In the latter case (`virtualenvs`), we aren't using it anymore either, but this class is fairly trivial and any third-party users who depended on it can easily roll their own.

### Refactoring of `VirtualEnv`

The `VirtualEnv` class was a bit of a mess for awhile now, thanks to new features being added for the last few years. Specifically, the addition of `pyenv` support made for some awkward method overrides. When using `pyenv`, the Jenkins experience is pretty awful because each `run` command generates ~70 lines of logging output just for `pyenv` boilerplate. Furthermore, the interface wasn't great because we modified the script command in order to source the `activate` script.

After a lot of investigation, we've determined that really the only thing that Jenkins cares about as far as creating a virtualenv is concerned is the `PATH`. So long as the virtualenv's paths appear ahead of the system ones, everything should be fine. The `activate` script also seems to set a `VIRTUAL_ENV` var, but as far as we can see, it's only used internally for other places in `activate`. However, if we find that other Python tools depend on this environment variable, we can easily support it.

This means that we can finally use `withEnv` instead of modifying the script command, which means that we can support a superior interface similar to the `docker` singleton:

```groovy
virtualenv.create('python3').inside {
  sh 'myscript.py'
  sh 'myotherscript.py'
}
```

This means that Jenkinsfiles can finally use a single closure call to run as many commands as they like, whereas previously one needed to make multiple calls to `venv.run`.


nre-ableton:
Note to reviewers: I've tested this internally on a project of ours that uses a combination of pyenv on Unix platforms and regular virtualenvs on Windows. Also, as a sanity check, I tested this in our sandbox environment with this script:

```groovy
library 'python-utils@nre/master/4654-simpler-virtualenvs'

node('generic-linux') {
    virtualenv.create('python3').inside {
        sh 'which python'
        sh 'python --version'
    }
}
```

When run, it prints `/jenkins/workspace/Utilities/sandbox/.venv/venv-92317815/bin/python` and `Python 3.8.10`.

mst-ableton:
There's a lot here, I'll have to take a look tomorrow...

src/com/ableton/Pyenv.groovy
```diff
@@ -8,6 +8,28 @@ package com.ableton
...
+   */
+  String pythonVersion
+  /**
+   * Pyenv root directory.
+   * <strong>Required value, may not be null!</strong>
+   */
+  String pyenvRoot
+
+  Pyenv(Object script, String pythonVersion, String pyenvRoot) {
+    this.script = script
```
mst-ableton, line 35: Shouldn't we assert that the values passed in aren't null here?

vars/pyenv.groovy
```diff
@@ -5,10 +5,21 @@ import com.ableton.VirtualEnv
 /**
  * Create a new {@link com.ableton.VirtualEnv} instance using pyenv.
  * @param pythonVersion Python version, as given by pyenv versions --list.
- * @param pyenvRoot Path of the pyenv installation. If {@code null}, then the pyenv root
- *        will be detected from the environment in {@code PYENV_ROOT}.
```
mst-ableton: Isn't this still the case? Shouldn't we leave the documentation in?

mst-ableton (commented):
Just a few questions, otherwise it looks good.

src/com/ableton/Pyenv.groovy
nre-ableton: No, we don't assert in constructors. I don't think we've formally documented this practice, but I don't think it's a good practice since most users don't expect creation of a new object to fail and seldom surround it with `try/catch`. We assert these values at usage.

vars/pyenv.groovy
nre-ableton: I guess it's the case that the pyenv root needs to be specified somehow, but it's not the case that this parameter exists anymore. :smile: Anyways, you're right, we should say _something_. Fixed in b923a9f.

mst-ableton (approved):
lgtm!

nre-ableton:
@ablbot autosquash

src/com/ableton/Pipenv.groovy
```diff
@@ -1,54 +0,0 @@
-package com.ableton
```
ala-ableton, line 1: Please reword the commit message of caf0c2d05158d126ee013e34e751b65ee8e7c9da to remove all subjective views on pipenv. We don't use it anymore, but that doesn't mean anything about the project itself.

> current pipenv releases (if there are any).

There are many: https://github.com/pypa/pipenv/releases

src/com/ableton/VirtualEnv.groovy
```diff
@@ -126,6 +132,15 @@ class VirtualEnv implements Serializable {
     script.dir(venvRootDir) { script.deleteDir() }
   }
 
+  /**
+   * Run a closure body inside of the virtualenv.
+   *
+   * @param body Closure body to execute.
+   */
+  void inside(Closure body) {
+    script.withEnv(["PATH+EXTRA=${venvBinDir}"]) { body() }
```
ala-ableton, line 89: We can write whatever we want after `PATH+` so let's use something that makes a bit more sense (like `VIRTUALENV` or `VENVBIN`).

src/com/ableton/VirtualEnv.groovy
```diff
@@ -149,20 +149,10 @@ class VirtualEnv implements Serializable {
...
-    String scriptCommand = """
-      ${this.activateCommands}
-      ${arguments.script}
-    """
-    // We shouldn't modify the original arguments map
-    Map newArguments = arguments.clone()
-    // Replace the original `script` command with the venv-activated one
-    newArguments['script'] = scriptCommand
-    newArguments['label'] = arguments.label ?:
-      "Run command in virtualenv: ${arguments.script}"
```
ala-ableton, line 149: Why are we losing the `label` in 7ebbf7202ef60be4de9ea2299ac34fb5545e7116?

src/com/ableton/VirtualEnv.groovy
```diff
@@ -171,13 +161,7 @@ class VirtualEnv implements Serializable {
    * @param command Command to run.
    */
   void run(String command) {
-    script.sh(
-      label: "Run command in virtualenv: ${command}",
```
ala-ableton, line 160: Why are we losing the `label` in 7ebbf7202ef60be4de9ea2299ac34fb5545e7116?

src/com/ableton/Pyenv.groovy
```diff
@@ -0,0 +1,50 @@
+package com.ableton
+
+
+/**
+ * Uses a pyenv (which must be installed on the node) to install a specific Python version
```
ala-ableton, line 5: "a pyenv" sounds a bit weird.

ala-ableton (changes requested):

src/com/ableton/Pipenv.groovy
nre-ableton: See 49b0b28.

src/com/ableton/VirtualEnv.groovy
nre-ableton: Where is the documentation on `PATH+WHATEVER`?

ala-ableton: https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#withenv-set-environment-variables

nre-ableton: Fixed in a153738.

src/com/ableton/Pipenv.groovy
ala-ableton: Thanks

src/com/ableton/Pyenv.groovy
nre-ableton: Fixed in 013f6db.

src/com/ableton/VirtualEnv.groovy
nre-ableton: Fixed in 750094b.

ala-ableton (commented):
Please rebase/squash and I'll have another look on Monday. Thanks!

src/com/ableton/VirtualEnv.groovy
```diff
@@ -149,20 +149,15 @@ class VirtualEnv implements Serializable {
    */
   @SuppressWarnings('MethodReturnTypeRequired')
   def run(Map arguments) {
-    assert arguments
-    assert arguments.containsKey('script')
```
ala-ableton, line 138: Why not assert anymore?

src/com/ableton/Pyenv.groovy
```diff
@@ -19,7 +41,10 @@ class Pyenv implements Serializable {
    * @param pyenvRoot Path to the installation of pyenv.
    * @return New instance of VirtualEnv object.
    */
-  static VirtualEnv createVirtualEnv(Object script, String python, String pyenvRoot) {
```
ala-ableton: In 3c18137f8ffb40f97e933e8237645ebc68c7e896, the docstrings for the parameters need to be removed.

src/com/ableton/Pyenv.groovy
```diff
@@ -8,6 +8,28 @@ package com.ableton
  * @see com.ableton.VirtualEnv
  */
 class Pyenv implements Serializable {
+  /**
+   * Script context.
+   * <strong>Required value, may not be null!</strong>
+   */
+  Object script
+  /**
+   * Python version, as listed in {@code pyenv install --list}.
```
ala-ableton, line 17: Line 40 says 
https://github.com/Ableton/python-pipeline-utils/blob/3c18137f8ffb40f97e933e8237645ebc68c7e896/src/com/ableton/Pyenv.groovy#L40

src/com/ableton/Pyenv.groovy
```diff
@@ -24,6 +24,13 @@ class Pyenv implements Serializable {
    */
   String pyenvRoot
 
+  Pyenv(Object script, String pythonVersion) {
+    this.script = script
+    this.pythonVersion = pythonVersion
+    assert script.env['PYENV_ROOT']
```
ala-ableton, line 30: This contradicts https://github.com/Ableton/python-pipeline-utils/pull/102#discussion_r970693413, where you wrote "No, we don't assert in constructors."

vars/pyenv.groovy
```diff
@@ -4,11 +4,27 @@ import com.ableton.VirtualEnv
...
+ * {@link #createVirtualEnv(String, String)} function instead.
+ *
  * @param pythonVersion Python version, as given by pyenv versions --list.
- * @param pyenvRoot Path of the pyenv installation. If {@code null}, then the pyenv root
- *        will be detected from the environment in {@code PYENV_ROOT}.
+ * @param pyenvRoot Path of the pyenv installation.
  * @return New {@link com.ableton.VirtualEnv} instance.
  */
-VirtualEnv createVirtualEnv(String pythonVersion, String pyenvRoot = null) {
-  return new Pyenv(this, pythonVersion, pyenvRoot ?: env.PYENV_ROOT).createVirtualEnv()
```
ala-ableton: I don't really see the value of 99e07088674a454670c8fbc4ead269cc23116eb9. It only seems to add more code.

ala-ableton (changes requested):

src/com/ableton/VirtualEnv.groovy
nre-ableton: My original reasoning was that it wasn't necessary because we were just passing the arguments map along to `sh`, but since you convinced me to re-add the custom label, I guess we should probably return it. :smile: Fixed in 95fc247.

src/com/ableton/Pyenv.groovy
nre-ableton: I'll take care of this during rebase.

src/com/ableton/Pyenv.groovy
nre-ableton: Fixed in 8a3e15f.

vars/pyenv.groovy
nre-ableton: Ok, we can remove it if you like. See 012ed7a.

ala-ableton (commented):
Please rebase/squash while addressing https://github.com/Ableton/python-pipeline-utils/pull/102#discussion_r980284391, then I'll have another look. Thanks!

vars/pyenv.groovy
```diff
@@ -4,11 +4,11 @@ import com.ableton.VirtualEnv
...
- * @param python Python version, as given by pyenv versions --list.
+ * @param pythonVersion Python version, as given by pyenv versions --list.
  * @param pyenvRoot Path of the pyenv installation. If {@code null}, then the pyenv root
  *        will be detected from the environment in {@code PYENV_ROOT}.
  * @return New {@link com.ableton.VirtualEnv} instance.
  */
-VirtualEnv createVirtualEnv(String python, String pyenvRoot = null) {
-  return Pyenv.createVirtualEnv(this, python, pyenvRoot ?: env.PYENV_ROOT)
+VirtualEnv createVirtualEnv(String pythonVersion, String pyenvRoot = null) {
+  return new Pyenv(this, pythonVersion, pyenvRoot ?: env.PYENV_ROOT).createVirtualEnv()
```
ala-ableton, line 13: I don't mind creating a `Pyenv` object, but that object should be able to create any virtualenv, not just ones that use this specific `pythonVersion`. Said differently: `python`/`pythonVersion` should still be an argument of `createVirtualEnv`, like in `VirtualEnv.create`.

ala-ableton: `pyenvRoot` does make sense as a parameter to the constructor of `Pyenv`.

nre-ableton: True. Fixed in 7b6ad08.

ala-ableton (commented):
The fixups look good. Please rebase/squash and I'll do a final review. Thanks!

ala-ableton (changes requested):
Once last issue to address, then LGTM 👍🏻

src/com/ableton/Pyenv.groovy
```diff
@@ -16,10 +32,12 @@ class Pyenv implements Serializable {
    * @param script Script context.
    *               <strong>Required value, may not be null!</strong>
    * @param python Python version, as given by pyenv versions --list.
```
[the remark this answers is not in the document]
nre-ableton: Fixed in 624effd.

ala-ableton (approved):
🎉

nre-ableton:
@ablbot rebase

nre-ableton:
@ala-ableton I totally forgot about documentation! The README was quite outdated. Please have another look.

ala-ableton (approved):

nre-ableton:
@ablbot autosquash

PR #105, opened 2022-11-29
Define VIRTUAL_ENV environment variable

nre-ableton:
When we implemented `VirtualEnv.inside` in 7cd11f9, we thought that the
`VIRTUAL_ENV` environment variable was only used by the `activate` and
`deactivate` scripts, and therefore it would be safe to only set the
`PATH` variable.

However, we have encountered at least one piece of software which relies
on this variable being set, and there may be others out there as well.
In order to keep compatibility with them, we should define this
variable.


mst-ableton (approved):
lgtm

PR #121, opened 2023-08-10
Use generic Python 3 for VirtualEnv.create test

nre-ableton:
We don't need to specify a specific Python 3.x version, we can just rely
on the host to give us some Python 3. This is just a sanity test, after
all.


anz-ableton (approved):
LGTM.

PR #124, opened 2023-10-24
Fail if pyenv could not install the Python version

nre-ableton:
Fixes https://github.com/AbletonAppDev/devtools/issues/5558


src/com/ableton/Pyenv.groovy
```diff
@@ -63,7 +63,9 @@ class Pyenv implements Serializable {
       // If we failed to create the virtualenv, test to see if the requested Python
       // version is supported. We don't do this pre-emptively to spare some work in the
       // pipeline, but if the operation failed, it is nice to fail with a clear error.
-      if (!versionSupported(pythonVersion)) {
+      if (versionSupported(pythonVersion)) {
+        script.error "Installation of Python ${pythonVersion} failed with code ${result}"
```
rco-ableton, line 67: Does `script.error` act like a `return`, in which case we can drop the `else` and save a level of indentation...

nre-ableton: Under the hood it raises an exception, but this isn't obvious to either humans or code analysis tools.

rco-ableton (approved):

nre-ableton:
@ala-ableton Are you interested in reviewing this change?

ala-ableton (approved):

PR #126, opened 2023-11-21
Update to Gradle 8.4

nre-ableton:


nre-ableton:
> (I didn't review the `gradlew` stuff as I presume it's auto-generated, right?)

Yes, that's correct.

PR #144, opened 2024-11-06
Retry python installation

anz-ableton:
The Python installation is now retried up to three times. This should
help to mitigate failed installations due to temporary unavailable
dependency sources.
Calling the Pipelines error method will cancel the retry loop, this is
why an exception is now thrown inside the retry block instead.


src/com/ableton/Pyenv.groovy
```diff
@@ -69,10 +73,11 @@ class Pyenv implements Serializable {
...
       """,
-    )
-
-    if (result != 0) {
-      script.error "Installation of Python ${pythonVersion} failed with code ${result}"
+      )
+      if (result != 0) {
+        throw new Exception("Installation of Python ${pythonVersion} failed with code " +
+          "${result}")
+      }
```
nre-ableton, line 80: Actually, let's get rid of this entirely. We can drop `returnStatus: true` in the `sh` command and just have it be alone in the `retry` loop with nothing else.

nre-ableton (changes requested):
Also some nitpicks for the commit message of 8932ef027511158d6a306f62094a18df268aeb4c:

> The Python installation is now retried up to three times. This should
help to mitigate failed installations due to temporarily unavailable
sources for dependencies.
> Calling the pipeline's error method will cancel the retry loop, this is
why an exception is now thrown inside the retry block instead.

...however, you should probably drop the second sentence if we remove `returnStatus` in `sh`.

src/com/ableton/Pyenv.groovy
anz-ableton: I like the idea. See 64d44acb31ad027322529a298d600f47473b4e41 which also removes the second sentence from the commit message.

src/com/ableton/Pyenv.groovy
```diff
@@ -60,9 +60,9 @@ class Pyenv implements Serializable {
 
     VirtualEnv venv = new VirtualEnv(script, randomSeed)
     script.retry(INSTALLATION_RETRIES) {
-      int result = venv.script.sh(
+      venv.script.sh(
         label: "Install Python version ${pythonVersion} with pyenv",
-        returnStatus: true,
+        returnStatus: false,
```
nre-ableton, line 65: This line should just be removed, as that is the default behavior.

nre-ableton (changes requested):

src/com/ableton/Pyenv.groovy
anz-ableton: See 336937b52a634b5d881393e39164c67840286214. I also addressed the linter complaints with this commit.

src/com/ableton/Pyenv.groovy
```diff
@@ -8,6 +8,9 @@ package com.ableton
  * @see com.ableton.VirtualEnv
  */
 class Pyenv implements Serializable {
+
```
nre-ableton, line 11:
```suggestion
```

src/com/ableton/Pyenv.groovy
```diff
@@ -8,6 +8,9 @@ package com.ableton
...
+  static final int INSTALLATION_RETRIES = 3
```
nre-ableton, line 12:
```suggestion
  static private final int INSTALLATION_RETRIES = 3
```

nre-ableton (changes requested):
Almost there! Just a few nits, and a typo-ish in the commit message of 336937b52a634b5d881393e39164c67840286214 ("...due to _temporarily_ unavailable...")

anz-ableton:
> Almost there! Just a few nits, and a typo-ish in the commit message of [336937b](https://github.com/Ableton/python-pipeline-utils/commit/336937b52a634b5d881393e39164c67840286214) ("...due to _temporarily_ unavailable...")

What do you think of:

> ... This should help to mitigate failed installations caused by dependency sources being unavailable temporarily.

nre-ableton:
> What do you think of:
> 
> > ... This should help to mitigate failed installations caused by dependency sources being unavailable temporarily.

Sounds good, but "temporarily unavailable" reads better than "unavailable temporarily", since you asked. :sweat_smile:

src/com/ableton/Pyenv.groovy
```diff
@@ -43,11 +46,23 @@ class Pyenv implements Serializable {
...
+
     VirtualEnv venv = new VirtualEnv(script, randomSeed)
-    int result = venv.script.sh(
-      label: "Install Python version ${pythonVersion} with pyenv",
-      returnStatus: true,
-      script: """
+    script.retry(INSTALLATION_RETRIES) {
+      venv.script.sh(
+        label: "Install Python version ${pythonVersion} with pyenv",
+        script: """
```
nre-ableton, line 65: I think the indent level is off for the below lines, sorry for not noticing that before.

nre-ableton (changes requested):

src/com/ableton/Pyenv.groovy
anz-ableton: No worries, see 5a5876fb3a09e2f0ba02dfc9f64487f1633198da.

nre-ableton:
@ablbot autosquash

nre-ableton (commented):
LGTM pending rebase and a version bump

nre-ableton (approved):

PR #145, opened 2024-11-08
Trim the pythonVersion

anz-ableton:
This parameter is likely to contain trailing newlines, e.g. when read from a Python version file.


anz-ableton:
There's another patch (32f28520c37e00fd63a02a4db19c5d136a8931e0) to satisfy groovylint.

ala-ableton (approved):

PR #177, opened 2025-08-19
Overwrite PATH rather than prepending to it

nre-ableton:
This fixes an error when installing Python 2.7.x via Pyenv, because paths like `/opt/local/bin` get in the way and cause problems. See discussion in https://github.com/pyenv/pyenv/issues/3281#issuecomment-3046275732.


src/com/ableton/Pyenv.groovy
```diff
@@ -122,15 +122,15 @@ class Pyenv implements Serializable {
...
+        "export PATH=\$PYENV_ROOT/bin:/usr/bin:/bin",
         'eval "\$(pyenv init --path)"',
         'eval "\$(pyenv init -)"',
       ]
     } else {
       String posixPyenvRoot = pyenvRoot[1] == ':' ?
         "/${pyenvRoot[0].toLowerCase()}/${pyenvRoot.substring(3)}" : pyenvRoot
       commands.add(
-        "export PATH=${posixPyenvRoot}/shims:${posixPyenvRoot}/bin:\$PATH"
+        "export PATH=${posixPyenvRoot}/shims:${posixPyenvRoot}/bin:/usr/bin:/bin"
```
ala-ableton, line 133: What is in `/usr/bin` and `/bin` on Windows?

nre-ableton: The contents of `C:\Program Files\git\usr\bin` and `C:\Program Files\git\bin`, respectively. Those directories seem to have the same contents, FWIW.

src/com/ableton/Pyenv.groovy
```diff
@@ -126,7 +126,11 @@ class Pyenv implements Serializable {
       commands.add("export PATH=${posixPyenvRoot}/shims:${posixPyenvRoot}/bin:\$PATH")
     } else {
       commands += [
-        "export PATH=\$PYENV_ROOT/bin:\$PATH",
+        // Unlike on Windows, we can't prepend to the $PATH because this causes problems
+        // on macOS with some Python versions, especially when other Python versions have
+        // been installed via MacPorts/Homebrew. Fortunately, we shouldn't need anything
+        // other than binaries in /usr/bin and /bin to build Python if necessary.
+        "export PATH=\$PYENV_ROOT/bin:/usr/bin:/bin",
```
ala-ableton, line 133: `/usr/sbin` and `/sbin` are part of the default `PATH`. Would it make sense to also have them here?

